PR #257, opened 2025-02-03
feat: add db cipher env variable

### LINKED ISSUE #33: SPIKE: Figure out path forward on Philadelphia database connection
### Background
Philly is having challenges connecting their on-prem SQL server to the eCR Viewer. Specifically, it seems their SQL Server requires some kind of cipher that our application doesn't support today. Here's what we know:

```
Philly notes:
- SQLSERVER version: unknown
- They have connected to the database on an ec2 instance from within the AWS network setup for dibbs

Philly db security settings
System default sect
CipherString = HIGH
MinProtocol = TLSv1.2
SecLevel = 0

Asks:
1. Can the app accept an environment variable that sets the cipher for the app client
	- What client does the app use?
	- How do we set the Cipher on the client?

Issues:
1. Postgres connection fails then the container rolls back. Check on the security groups.
1. Check the tech acceptance docs, remove `.json` from URL
1. Check with the team on how current these docs are: https://cdcgov.github.io/phdi/latest/containers/orchestration.html#operation/upload_config_configs__processing_config_name__put
```

### Request
We need to figure out how to make this work for Philadelphia. Either we'll have to do as they suggest and set a cipher on the Viewer (somehow) or come up with another way for them to connect.

### POC
Alis has been the POC for the Philly engagement, and they can direct you towards additional information on the Philly side if needed.

BobanL:
# PULL REQUEST

## Summary

- Add env variable for DB cipher

## Related Issue

Fixes #33 

## Acceptance Criteria

## Additional Information

- I was unable to emulate sql server 2016 with the different ciphers, but I was able to change the cipher
- Cipher is a string of upper cased ciphers separated by `:`.
  - [Tedious Docs](https://tediousjs.github.io/tedious/api-connection.html#:~:text=options.cryptoCredentialsDetails) - [NodeJS Docs](https://nodejs.org/docs/latest/api/tls.html#modifying-the-default-tls-cipher-suite)


robertmitchellv (approved):
🎉 👍🏻

PR #266, opened 2025-02-03
perf: make ecr viewer page ssr

### LINKED ISSUE #56: Access database directly
## What needs to be done

To retrieve the FHIR bundle to be displayed in the `view-data` page of the eCR-Viewer we make an API call. However if we separated the code around handling the API request and response from the code that queries our databases there is no reason why the app could not directly call our data retrieval call directly.  

## Why it needs to be done

This would remove the latency of making an API call, and simplify some of the code.

## Acceptance Criteria

- the `view-data` page can retrieve the FHIR bundle without making an API query.

mcmcgrath13:
# PULL REQUEST

## Summary

Push as much as we reasonable can (without too big of a change/refactor) of the ecr viewer `view-data` page to the server side.

Highlights:
* Make the `page.tsx` a server side component. Use the `get_fhir_data` function to directly call the DB instead of an http API call
* Audit all `"use client"` directives and do some splitting to make only the components that need to be client side, client side
* Directly use a loading state on the side nav as it is slow and we don't need to block the whole page on that (and make the markup actually match)
* Make the `ToolTipElement` only render on the client side (otherwise there are some hydration issues)
* Use the `node-html-parser` package for table wrangling instead of the `DOMParser` so this can be done on the server
* Add an `Icon` component which wraps the uswds version and only brings in what we need (weird issues with client bundling otherwise)
* 

## Related Issue

Fixes #56

## Acceptance Criteria

## Additional Information

Performance profile test results with 4x CPU throttling and fast 4G network speeds:

Before:
image

After:
image

### DIFF
--- a/containers/ecr-viewer/src/app/view-data/page.tsx
+++ b/containers/ecr-viewer/src/app/view-data/page.tsx
@@ -1,22 +1,21 @@
-import AccordionContent from "@/app/view-data/components/AccordionContent";
 import React from "react";
 import { Bundle } from "fhir/r4";
-import { PathMappings } from "./utils/utils";
-import SideNav from "./components/SideNav";
-import { ExpandCollapseButtons } from "@/app/view-data/components/ExpandCollapseButtons";
-import { EcrLoadingSkeleton } from "./components/LoadingComponent";
-import RetrievalFailed from "./retrieval-failed";
-import { ECRViewerLayout } from "./components/ECRViewerLayout";
-
 import { Grid, GridContainer } from "@trussworks/react-uswds";
 
-import EcrSummary from "./components/EcrSummary";
 import { get_fhir_data } from "../api/fhir-data/fhir-data-service";
 import {
   evaluateEcrSummaryConditionSummary,
   evaluateEcrSummaryEncounterDetails,
   evaluateEcrSummaryPatientDetails,
 } from "../services/ecrSummaryService";
+import { PathMappings } from "./utils/utils";
+import AccordionContent from "@/app/view-data/components/AccordionContent";
+import { EcrLoadingSkeleton } from "./components/LoadingComponent";
+import { ECRViewerLayout } from "./components/ECRViewerLayout";
+import { ExpandCollapseButtons } from "@/app/view-data/components/ExpandCollapseButtons";
+import EcrSummary from "./components/EcrSummary";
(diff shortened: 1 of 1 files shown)


containers/ecr-viewer/seed-scripts/baseECR/.gitignore
```diff
@@ -1,2 +1,2 @@
 bundle.json
-LA
\ No newline at end of file
+**/*.xml
```
JNygaard-Skylight, line 2: Did you mean to keep this in?

JNygaard-Skylight (changes requested):

containers/ecr-viewer/src/app/view-data/components/ToolTipElement.tsx
BobanL, line 1: Do we need both `"use client"` and `ForceClient` in the tool tip element?

containers/ecr-viewer/seed-scripts/baseECR/.gitignore
mcmcgrath13: yeah. It felt like a safe thing to do at this moment in time

containers/ecr-viewer/src/app/view-data/components/ToolTipElement.tsx
mcmcgrath13: yes as the file uses react hooks, we need "use client", then the ForceClient makes sure we don't end up with weird hydration errors

containers/ecr-viewer/src/app/view-data/page.tsx
```diff
@@ -1,67 +1,63 @@
...
 import { ECRViewerLayout } from "./components/ECRViewerLayout";
 
-interface FetchError {
-  status: number;
-  message: string;
-}
+import { Grid, GridContainer } from "@trussworks/react-uswds";
+
+import EcrSummary from "./components/EcrSummary";
+import { get_fhir_data } from "../api/fhir-data/fhir-data-service";
```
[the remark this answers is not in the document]
mcmcgrath13: there wasn't, but now I've gone through and ordered them in a more typical
```
// react first
// other 3rd party...
//
// ours...
```
fashion

JNygaard-Skylight (approved):

PR #281, opened 2025-02-10
docs: update readme badges

BobanL:
# PULL REQUEST

## Summary

- Update the badges on the readme with links to dibbs-ecr-viewer rather than phdi

## Additional Information

- The readme will require an entire rewrite to reflect dibbs-ecr-viewer, but for now this will just update those badgets


JNygaard-Skylight (approved):
LGTM!

PR #306, opened 2025-02-13
fix: "eCR Library" should be a header to support screen readers

### LINKED ISSUE #288: [A11y Library Fix] - "eCR Library" should be a header
## What needs to be done

Change the "eCR Library" text to be an HTML header element (it's currently a span).

## Why it needs to be done

This is an accessibility issue - screenreader users would not "see" the eCR Library heading in their content scan.

See Axe scan issue: https://cdc-devtools.dequecloud.com/issues/972b0840-6fd7-4f34-a4fd-f44ba2aba632

## Acceptance Criteria

- [ ] "eCR Library" element is a `h2` element

robertmitchellv:
# PULL REQUEST

## Summary

In order to ensure screen readers can understand the hierarchy of the information, the "eCR Library" text is changed to an `<h2>` tag, which will be under the `<h1>` tag "DIBBs eCR Viewer".

## Related Issue

Fixes #288 
- #288 

## Acceptance Criteria

- [x] "eCR Library" element is a `h2` element


containers/ecr-viewer/src/app/page.tsx
```diff
@@ -51,7 +51,7 @@ const HomePage = async ({
       <Header />
       <main className="overflow-auto height-full">
         <div className="margin-x-3 padding-y-105 display-flex flex-align-center">
-          <span className="text-bold font-sans-xl">eCR Library</span>{" "}
+          <h2 className="text-bold font-sans-xl">eCR Library</h2>{" "}
```
BobanL, line 54: h2 has some some additional margin by default that we can remove with class `margin-bottom-0`. Also can we remove that random space at the end (`{" "}`)?
- 

image

BobanL (changes requested):

containers/ecr-viewer/src/app/page.tsx
robertmitchellv: absolutely!

robertmitchellv: added the class to `h2` and removed the empty space if you want to take a look again!

BobanL (approved):
🎉

PR #309, opened 2025-02-13
refactor:  Fix `returnTableFromJson` key anti-pattern

### LINKED ISSUE #26: Fix `returnTableFromJson` key anti-pattern

Fine:
```js
```

Not fine:
```js
```

`returnTableFromJson` currently adds a `key` on the item in isolation of the map. Try to fix this (it might be a deep rabbit hole, so time box).

Ideally, turn the function into a component `TableFromJson` or similar. This might run into issues with data availability checks. If so, consider adding a validity check before rendering instead of rendering and maybe getting nothing.

jakewheeler:
# PULL REQUEST

## Summary

The goal of this PR is to:
- Fix issue described in ticket: #26 
- Replace the `returnTableFromJson` function that returns `JSX.Element | undefined` with a `JsonTable` React component
- Perform data checks before attempting to render `JsonTable`
- Reorganize code where appropriate
- Improve/create tests for these changes
- Add other small improvements (more specific naming, more specific typing, etc.)

## Related Issue

Fixes #26 

## Acceptance Criteria

- [x] Application functionality remains unchanged

## Additional Information

It is difficult to tell from the snapshot test that changed, but the actual rendered HTML appears to be identical to what it was previously. Please feel free to double check me here.


containers/ecr-viewer/src/app/view-data/components/common.tsx
```diff
@@ -248,64 +245,25 @@ export const returnHtmlTableContent = (
   title: string,
   outerBorder = true,
   className = "",
-) => {
+): JSX.Element | undefined => {
```
jakewheeler, line 248: I think more can be done with `returnHtmlTableContent` but my main goal here was only to update it to work with the changes made to `JsonTable`. We can figure out what to do with this in a follow-up ticket.

containers/ecr-viewer/src/app/services/htmlTableService.tsx
```diff
@@ -80,7 +80,7 @@ export function formatTablesToJSON(htmlString: string): TableJson[] {
       li.querySelectorAll("table").forEach((table) => {
         tables.push(processTable(table));
       });
-      jsonArray.push({ resultId, resultName, tables });
+      jsonArray.push({ resultId: resultId ?? undefined, resultName, tables });
```
mcmcgrath13, line 83: could you move the `??` up to the `resultId` assignment line? more consistent with other sections here

Should/could `getDataId` be updated so it returns undefined directly?

containers/ecr-viewer/src/app/view-data/components/JsonTable.tsx
```diff
@@ -0,0 +1,61 @@
...
+        <tr key={`table-row-${i}`}>
+          {Object.values(row).map(({ value }, j) => (
+            <td key={`table-col-${j}`}>{value ?? noData}</td>
+          ))}
+        </tr>
+      ))}
+    </BaseTable>
+  );
+};
+function useConstructColumns(
```
mcmcgrath13, line 50:
```suggestion

function useConstructColumns(
```

containers/ecr-viewer/src/app/view-data/components/common.tsx
```diff
@@ -248,64 +245,25 @@ export const returnHtmlTableContent = (
...
-            ))}
-          </tr>
-        ))}
-      </BaseTable>
-    );
-  }
+  if (tableJson.length === 0) return undefined;
+
+  return (
+    <Fragment key={`${Math.random()}`}>
```
mcmcgrath13, line 255: this is a different ticket, but do we still need this `key`?

containers/ecr-viewer/src/app/services/htmlTableService.tsx
```diff
@@ -15,7 +15,7 @@ export interface TableRow {
   };
 }
 
-export interface TableJson {
+export interface HtmlTableJson {
```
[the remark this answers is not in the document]
mcmcgrath13: rename makes sense to me!

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
```diff
@@ -616,10 +616,14 @@ export const evaluateEncounterCareTeamTable = (
...
-  return returnTableFromJson(
-    { resultName: "Encounter Care Team", tables: [tables] },
-    true,
-    "caption-data-title margin-y-0",
+  if (!tables.length) return undefined;
+
+  return (
+    <JsonTable
+      jsonTableData={{ resultName: "Encounter Care Team", tables: [tables] }}
+      outerBorder
```
mcmcgrath13, line 624:
```suggestion
```

containers/ecr-viewer/src/app/services/socialHistoryService.tsx
```diff
@@ -60,9 +60,13 @@ export const evaluateTravelHistoryTable = (
...
-  return returnTableFromJson(
-    { tables, resultName: "Travel History" },
-    true,
-    "caption-data-title margin-y-0",
+  if (!tables.flat().length) return undefined;
+
+  return (
+    <JsonTable
+      jsonTableData={{ tables, resultName: "Travel History" }}
+      outerBorder
```
mcmcgrath13, line 68:
```suggestion
```

containers/ecr-viewer/src/app/tests/components/JsonTable.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+    expect(screen.getByText("col1")).toBeInTheDocument();
+    expect(screen.getByText("col2")).toBeInTheDocument();
+    expect(screen.getByText("val1")).toBeInTheDocument();
+    expect(screen.getByText("val2")).toBeInTheDocument();
+  });
+
+  it("returns nothing if required table data is not available", () => {
+    const { container } = render(<JsonTable jsonTableData={{}} />);
+    expect(container).toBeEmptyDOMElement();
+  });
```
mcmcgrath13, line 29: could we add a test for returning nothing if the table data is like medium available but not actually? (has name and after flattening will be empty)

containers/ecr-viewer/src/app/tests/components/common.test.tsx
```diff
@@ -110,5 +84,16 @@ describe("common tests", () => {
       ).toBeInTheDocument();
       expect(screen.getAllByText("2000-02-04T21:02:00.000Z")).toHaveLength(2);
     });
+
+    it("returns nothing if required table data is not available", () => {
+      const result = returnHtmlTableContent(
+        undefined as any,
```
mcmcgrath13, line 90: same here on the medium unavailable data

containers/ecr-viewer/src/app/view-data/components/JsonTable.tsx
```diff
@@ -0,0 +1,61 @@
+import { HtmlTableJson, TableRow } from "@/app/services/htmlTableService";
```
mcmcgrath13, line 1: should the `TableRow` also be changed to `HtmlTableRow` or `HtmlTableJsonRow` or something?

containers/ecr-viewer/src/app/services/htmlTableService.tsx
jakewheeler: I think `getDataId()` can be left alone. I noticed `elem.getAttribute("data-id")` returns `string | null | undefined` so even if I were to change the `return null` to `return undefined` in the `else` block, it wouldn't make any difference. 

We could add some logic around `elem.getAttribute("data-id")` but handling this issue during the `resultId` assignment feels like a simpler way to go.

containers/ecr-viewer/src/app/view-data/components/JsonTable.tsx
jakewheeler: Yes, great suggestion!

containers/ecr-viewer/src/app/tests/components/JsonTable.test.tsx
jakewheeler: I added an additional test here and another for `returnHtmlTableContent` that have partial data available to them. Let me know what you think!

mcmcgrath13 (approved):
thanks!

PR #310, opened 2025-02-13
refactor: moar server side things

mcmcgrath13:
# PULL REQUEST

## Summary

Refactor things so that there's a clearer split between client and server side code. With these changes, we should never need to ship/run `fhirpath` and friends on the client. This almost halves our js bundle size.

Highlights:
* Spin off from `formatService`: `format-utils`, `htmlTableService`, `formatDateService`
  * This allows better tree shaking as we now have tighter third party imports based on what is actually being done in each case
* Create a `utils` folder and put our util code there
* Make `Patient Banner` a more layout/styling component and pass in the data fully evaluated instead of doing evaluation there
* Split off `EvaluateTableRow` into its own client side component and keep the data evaluation/munging in the server side `EvaluateTable` component
* Split off `ECRTableClientBase` for the server-side-able pieces of the table
* Enforce key files stay server only via `server-only` package/import

## Additional Information

Bundle size data:

Before:
image
image

After:
<image width="427" src="https://github.com/user-attachments/assets/f4ca7a1e-b56a-4bb8-a59d-7e6c69aac388" />
image

Network Requests:

Dev mode:
Before:
image

After:
image

Built:
Before:
image

After:
image

## Checklist


containers/ecr-viewer/src/app/view-data/components/EvaluateTableRow.tsx
```diff
@@ -0,0 +1,72 @@
...
+}: {
+  tableRowData: TableRowData;
+  numCols: number;
+}) => {
+  const [hiddenComment, setHiddenComment] = useState(true);
+  const id = useId();
+
+  const { rowCellsData, hiddenRow } = tableRowData;
+
+  if (rowCellsData.length === 0) return <></>;
```
mcmcgrath13, line 33:
```suggestion
  if (rowCellsData.length === 0) return null;
```

jakewheeler (approved):
Great job, Mary! Your walkthrough was very helpful.

PR #318, opened 2025-02-14
refactor: clean up Accordion usage

### LINKED ISSUE #317: Remove `AccordionContainer` wrapper component
## What needs to be done

Currently, the only usage of `AccordionContainer` is within the `AccordionContent` component. `AccordionContent` passes along its `AccordionItems` to `AccordionContainer`, which modifies them, and finally passes them to the USWDS `Accordion` component where the items are rendered.

We should remove the middle step of passing the items to the `AccordionContainer` and instead have `AccordionContent` pass its items directly to the USWDS `Accordion` component.

## Why it needs to be done

As you can probably tell from the description, we have a lot of accordion-related component nesting that can be simplified. This change should help to reduce the cognitive complexity when dealing with the primary accordion layout used by the `/view-data` page.

## Acceptance Criteria

- [ ] `AccordionContainer.tsx` is deleted
- [ ] `AccordionContainer.test.tsx` is deleted
- [ ] Functionality of `AccordionContainer` should be moved into `AccordionContent` (if keeping it is desired)
- [ ] `AccordionContainer` tests should be moved into `AccordionContent.test.tsx` and should continue to pass (if functionality remains)
- [ ] Accordion related types are cleaned up, if needed

## Technical Notes

N/A

jakewheeler:
# PULL REQUEST

## Summary

This PR aims to reduce code complexity by deleting the `AccordionContainer` component, which was a middle step between the `AccordionContent` component and the USWDS `Accordion` component. I've completed some other minor clean up as well.

## Related Issue

Fixes #317 

## Acceptance Criteria

- [x] `AccordionContainer.tsx` is deleted
- [x] `AccordionContainer.test.tsx` is deleted
- [x] Accordion related types are cleaned up, if needed

## Additional Information

Please see my comments for a couple of questions I have!


containers/ecr-viewer/src/app/view-data/components/AccordionContent.tsx
```diff
@@ -234,6 +244,36 @@ const AccordionContent: React.FC<AccordionContainerProps> = ({
...
+  index: number,
+): AccordionItemProps => {
+  return {
+    ...item,
+    id: `${
+      typeof item.title === "string" ? toKebabCase(item.title) : item.id
+    }_${index + 1}`,
+    title: <span>{item.title}</span>,
+  };
+};
```
jakewheeler, line 277: This functionality was moved over from `AccordionContainer` but I don't think we need it. Any reason not to delete this entirely?

containers/ecr-viewer/src/app/tests/components/AccordionContent.test.tsx
```diff
@@ -24,6 +24,16 @@ describe("Snapshot test for Accordion Content", () => {
...
     });
+
+    // This is an arbitrarily chosen test ID we expect one of the rendered accordion items to have.
+    // There is nothing significant about "encounter-info_2" in particular.
+    expect(
+      screen.getByTestId("accordionButton_encounter-info_2"),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByTestId("accordionItem_encounter-info_2"),
+    ).toBeInTheDocument();
```
jakewheeler, line 35: These tests can be deleted if we decide we don't need [what I pointed out here](https://github.com/CDCgov/dibbs-ecr-viewer/pull/318/files#r1960030295).

containers/ecr-viewer/src/app/view-data/components/AccordionContent.tsx
mcmcgrath13: I don't see why we'd need the `span` (especially without any styling on it). I do think the `id` handling is kind of nice

containers/ecr-viewer/src/app/view-data/components/AccordionContent.tsx
```diff
@@ -75,8 +79,9 @@ const AccordionContent: React.FC<AccordionContainerProps> = ({
     );
   };
 
-  const accordionItems: any[] = [
+  const accordionItems: AccordionItemProps[] = [
     {
+      id: "patient-info",
```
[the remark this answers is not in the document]
jakewheeler: I've added a slightly simpler postprocessing step to add that `id`. Let me know if you think this works 😄

containers/ecr-viewer/src/app/view-data/components/AccordionContent.tsx
```diff
@@ -229,11 +241,23 @@ const AccordionContent: React.FC<AccordionContainerProps> = ({
...
       ),
-      expanded: true,
-      headingLevel: "h3",
+      expanded,
+      headingLevel,
     },
   ];
 
-  return <AccordionContainer accordionItems={accordionItems} />;
+  const accordionItemsWithIds = accordionItems.map((item) => ({
```
mcmcgrath13, line 249: not 100% clear if this is the case, but if `expanded` and `headingLevel` are the same for every item, do we also want to just add them in this post-process mapping?

containers/ecr-viewer/src/app/tests/components/__snapshots__/AccordionContent.test.tsx.snap
```diff
@@ -86,20 +80,18 @@ exports[`Snapshot test for Accordion Content Given no data, info message for emp
       class="usa-accordion__heading"
     >
       <button
-        aria-controls="lab-info_4"
+        aria-controls="lab-info"
```
mcmcgrath13, line 83: should there be a corresponding `id` that this controls? Also, if so, I wonder if we need those numbers to avoid clashing with the headings ids?

containers/ecr-viewer/src/app/view-data/components/AccordionContent.tsx
jakewheeler: We could, but it's going to be a little awkward to do IMO. I had to create a new type, `AccordionItemWithoutId`, based on the `AccordionItemProps` type that allows us to create that `accordionItems` list without the `id` required by `AccordionItemProps`.

I think if we wanted to included `expanded` and `headingLevel` in the "building back up to the original type we want" process, we would need to update `AccordionItemWithoutId` to be something like `AccordionItemWithoutIdAndheadingLevelAndExpanded` 😆  or probably something named a bit better but hopefully you can see what I mean. Definitely open alternatives though if you have any suggestions.

containers/ecr-viewer/src/app/tests/components/__snapshots__/AccordionContent.test.tsx.snap
jakewheeler: You are correct! I just took a look into this and noticed that `AccordionContent.test.tsx` runs this code:
```ts
container.querySelectorAll("[id], [aria-describedby]").forEach((el) => {
  el.removeAttribute("id");
  el.removeAttribute("aria-describedby");
});
```
It looks like the reason this is here is to deal with tooltips having randomly generated `id`s and, I assume, causing problems for snapshots 😬 

FWIW I can see that the `div` in a real browser does have the corresponding expected `id`.
image

containers/ecr-viewer/src/app/view-data/components/AccordionContent.tsx
mcmcgrath13: if you do the `.map` in one line, I think it would work around the type issue

mcmcgrath13: e.g.
```
const accodionItems = [{starter items}, {...}].map(augment)
```

containers/ecr-viewer/src/app/tests/components/__snapshots__/AccordionContent.test.tsx.snap
mcmcgrath13: That random tooltip ID causes a lot of problems...

So it seems like these `aria-controls` are now more correct? is that the case?

containers/ecr-viewer/src/app/view-data/components/AccordionContent.tsx
jakewheeler: Yes, this should work. Thanks!

containers/ecr-viewer/src/app/tests/components/__snapshots__/AccordionContent.test.tsx.snap
jakewheeler: It should be as correct as it was before. The main difference here is really just the removal of the index number at the end of the `id` name. So now it's `clinical-info` instead of `clinical-info_3`.

mcmcgrath13: how does this interact with the side nav, which looks for the `toKebabCased` header to link an anchor to? Do we now have multiple things with the same ID?

jakewheeler: Apologies for my confusion here 🙂 

I've added the `span` with the `id` for the title back in that the side nav will link to. I've also modified the generated IDs that get applied to the other accordion item elements to once again include the `index + 1` at the end so that they don't conflict with the `id` of the `span`. 

Please let me know if you still have any other suggestions!

mcmcgrath13 (approved):
nice cleanup!

PR #319, opened 2025-02-14
Update DB schemas to use new ecr-viewer schema not dbo

### LINKED ISSUE #41: Schema updates for Philadelphia
### What is this?
Philly has requested that the eCR Viewer live in a separate schema from the default `DBO` schema, since everyone in their jurisdiction has access to that. They recommended an `eCR-Viewer` schema instead. We should be able to change this for all pilot partners, so can change the core and extended schema to accommodate this.

(Shouldn't need a migration script - this would be one of the first things that STLT partners run).

### Acceptance Criteria
- [ ] All tables, within both core and extended models, live in a custom `ecr-viewer` schema
- [ ] This new schema is created once on database initialization (run manually?)

gordonfarrell:
# PULL REQUEST

## Summary

Creates a new ecr-viewer schema for our tables to live in per Philly's request rather than landing in dbo by default

## Related Issue

Fixes #41 

## Acceptance Criteria

- [ ] All tables, within both core and extended models, live in a custom `ecr-viewer` schema
- [ ] This new schema is created once on database initialization (run manually?)

### DIFF
--- a/containers/ecr-viewer/sql/core.sql
+++ b/containers/ecr-viewer/sql/core.sql
@@ -1,6 +1,8 @@
 CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
 
-CREATE TABLE ecr_data (
+CREATE SCHEMA ecr_viewer;
+
+CREATE TABLE ecr_viewer.ecr_data (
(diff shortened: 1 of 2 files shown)


BobanL (approved):
Nice!

PR #320, opened 2025-02-14
fix: anchor tag css and sidenav activation handling

mcmcgrath13:
# PULL REQUEST

## Summary

Consistently account for the patient banner when handling anchor links as well as the side nav activation.

* Add a css var `--patient-banner-buffer` and set it in the layout based on the integrated viewer env var
* Use that css var in the `.nav-wrapper` class to avoid some in line classname logic
* Use that css var to add scroll margin to all potential link targets in the viewer
* Use that css var to create a tighter space for the intersection observer so it's less noisy (allows us to get rid of the flickery on click delay handler)

## Related Issue

Fixes #3198

## Additional Information

Before:
https://github.com/user-attachments/assets/548bfaab-eecd-4654-96e2-e5b07197158b

After:
https://github.com/user-attachments/assets/08522a99-f23b-4904-acc5-8d747bed793e


containers/ecr-viewer/src/app/layout.tsx
```diff
@@ -1,11 +1,19 @@
 import "../styles/styles.scss";
-import { PublicEnvScript } from "next-runtime-env";
+import { PublicEnvScript, env } from "next-runtime-env";
 
 export const metadata = {
   title: "DIBBs eCR Viewer",
   description: "View your eCR data in an easy-to-understand format.",
 };
 
+const PATIENT_BANNER_BUFFER = "3rem";
```
BobanL, line 9: The banner size is a little smaller than this at 2.75 rem. When I was playing around setting this to `2.75` and updating `scroll-margin-top` to add only `.5` looks a bit nicer with the spaces.

Before: (notice the blue bar up top)
image
After:
image

Before: (The section above is still slightly visible)
image

After:
image

mcmcgrath13: thanks! I'll give this a try. I did see the tiny bit of blue on the ecr summary link, but interestingly, locally, don't see that sliver of teal above demographics 🤔 - might be some browser/settings specific things

mcmcgrath13: P.S. example of gap between header teal banner and patient banner
image

BobanL: Yep this looks good! Thanks for the change :D

BobanL (approved):
👍

PR #328, opened 2025-02-18
refactor: have misc. notes use `JsonTable` - remove `addCaptionToTable`

### LINKED ISSUE #22: Fix `addCaptionToTable` key anti-pattern

Fine:
```js
```

Not fine:
```js
```

`addCaptionToTable` currently adds a `key` on the item in isolation of the map. Remove it (this one should be trivial).

mcmcgrath13:
# PULL REQUEST

## Summary

Instead of try to shove in a `caption` late in the processing chain, handle the "is this a string or is this a table" logic for misc. notes up front.

This allows us to remove `addCaptionToTable` entirely, thus removing the key antipattern. It also gets the dates in the table more consistently formatted with the rest of the app

## Related Issue

Fixes #22

## Acceptance Criteria

- [ ] `addCaptionToTable` should not require a hard coded `key`

## Additional Information

Before:
image

After:
image


BobanL (approved):
Looks good!

PR #330, opened 2025-02-19
fix: add "no eCR" message when no data listed

### LINKED ISSUE #35: Add "no eCRs found" message when everything is filtered out of Library
## What needs to be done

Working with design, add some kind of "no eCRs found" message to the eCR Library when filter/search are applied in such a way that nothing matches.
 
## Why it needs to be done

It's a confusing user experience to search or filter and get no results (especially because our page is so fast that it seems like it might have errored out).

## Acceptance Criteria

- Some kind of text/message is shown to the user when no results are returned for their query.

## Technical Notes
See screenshot of the current experience:
- 

## Additional context

Not sure if we already have designs for this? If not, please work directly with Ashton.

mcmcgrath13:
# PULL REQUEST

## Summary

When no eCR data is returned, show a message to that effect

## Related Issue

Fixes #35

## Acceptance Criteria

- [ ] Some kind of text/message is shown to the user when no results are returned for their query.

## Additional Information

image


mcmcgrath13:
> This looks good!
> 
> The only thing that I noticed (and this might be a design question) was that if someone searches using keywords and they don't show any results, it wasn't immediately obvious to me how to reset the Library (need to hover over the search bar -> find & click the clear button -> resubmit the search). Any opinions on this?

@ashton-skylight what do you think?

ashton-skylight:
> > This looks good!
> > The only thing that I noticed (and this might be a design question) was that if someone searches using keywords and they don't show any results, it wasn't immediately obvious to me how to reset the Library (need to hover over the search bar -> find & click the clear button -> resubmit the search). Any opinions on this?
> 
> @ashton-skylight what do you think?

Yeah, that's a great point. I wonder if a button reset near the text (No eCRs found...) text would be appropriate. I'll mock one up post-standup and send for feedback

mcmcgrath13:
> > > > This looks good!
> > > > The only thing that I noticed (and this might be a design question) was that if someone searches using keywords and they don't show any results, it wasn't immediately obvious to me how to reset the Library (need to hover over the search bar -> find & click the clear button -> resubmit the search). Any opinions on this?
> > > 
> > > 
> > > @ashton-skylight what do you think?
> > 
> > 
> > Yeah, that's a great point. I wonder if a button reset near the text (No eCRs found...) text would be appropriate. I'll mock one up post-standup and send for feedback
> 
> Thought about the reset a bit more - would prefer not to introduce more buttons that do the same function. With filtering, we have the reset button on the top and the search has a clear state. [Figma designs](https://www.figma.com/design/zAPyOEYwNtx3f2Y47aoq0G/eCR-Viewer-(Aug-2024---Present)?node-id=9775-8026&m=dev)
> 
> Instead, we could do one of two things:
> 
> 1. Have a general "No eCRs found" page for both filters and search (as shown by the attachment) with guiding copy to help the user
>    eCR Library - No eCRs found on filter and search
> 2. Create different messaging depending on the situation (1) user filters and search (2) user only filters (3) user only searches. In this case, the H1 and body text would be situation-specific
>    eCR Library - No eCRs found on search
> 
> eCR Library - No eCRs found on filter

I feel like this is adding quite a bit of complexity for a relatively straightforward situation (people are generally used to filters leading to no data). I do think it would be nice if hitting the `x` in the search actually submitted the clearing of the search instead of only clearing the text - thoughts on that?

ashton-skylight:
> I feel like this is adding quite a bit of complexity for a relatively straightforward situation (people are generally used to filters leading to no data). I do think it would be nice if hitting the `x` in the search actually submitted the clearing of the search instead of only clearing the text - thoughts on that?

Hmm that's a good idea. After comparing search bars across six applications, the accepted pattern is to use the 'x' to clear the text, and not the search entirely. If we did the clear search, when the user hovers over the 'x', would we be able to return a tooltip that tells them that they will clear the search before they complete the action?

mcmcgrath13:
> > I feel like this is adding quite a bit of complexity for a relatively straightforward situation (people are generally used to filters leading to no data). I do think it would be nice if hitting the `x` in the search actually submitted the clearing of the search instead of only clearing the text - thoughts on that?
> 
> Hmm that's a good idea. After comparing search bars across six applications, the accepted pattern is to use the 'x' to clear the text, and not the search entirely. If we did the clear search, when the user hovers over the 'x', would we be able to return a tooltip that tells them that they will clear the search before they complete the action?
> 
> Alternatively, many websites and applications allow you to return to the default state when clicking on the logo. We could provide that option when clicking on 'eCR library'.

The default state is restored when clicking on the logo. So maybe we're already good?

ashton-skylight:
> > > I feel like this is adding quite a bit of complexity for a relatively straightforward situation (people are generally used to filters leading to no data). I do think it would be nice if hitting the `x` in the search actually submitted the clearing of the search instead of only clearing the text - thoughts on that?
> > 
> > 
> > Hmm that's a good idea. After comparing search bars across six applications, the accepted pattern is to use the 'x' to clear the text, and not the search entirely. If we did the clear search, when the user hovers over the 'x', would we be able to return a tooltip that tells them that they will clear the search before they complete the action?
> > Alternatively, many websites and applications allow you to return to the default state when clicking on the logo. We could provide that option when clicking on 'eCR library'.
> 
> The default state is restored when clicking on the logo. So maybe we're already good?

Yeah, that works!

angelathe (approved):
🎉

PR #335, opened 2025-02-19
Fix spots I missed switching from DBO to new schema

gordonfarrell:
# PULL REQUEST

## Summary

Update a few queries to correctly reference new schema


mcmcgrath13 (approved):
can we make sure this gets caught when the e2e tests are back up and running?

thanks!

PR #336, opened 2025-02-19
refactor: `returnHtmlTableContent` key anti-pattern

### LINKED ISSUE #21: `returnHtmlTableContent` key anti-pattern

Fine:
```js
```

Not fine:
```js
```

`returnHtmlTableContent` currently adds a `key` on the item in isolation of the map. Try to fix this (it might be a deep rabbit hole, so time box).

Ideally, turn the function into a component `HtmlTable` or similar. This might run into issues with data availability checks. If so, consider adding a validity check before rendering instead of rendering and maybe getting nothing.

jakewheeler:
# PULL REQUEST

## Summary

This PR eliminates `returnHtmlTableContent` in favor of using the `JsonTable` component directly. This should help to simplify the code while also fixing the key anti-pattern described in the ticket.

I've also introduced `evaluatePlanOfTreatment()` to handle the `Plan of Treatment` piece of the overall `treatmentData` evaluation. This will check to see if data is available before attempting to construct table components.

## Related Issue

Fixes #21 

## Acceptance Criteria

- [x] Should not introduce any visual changes


mcmcgrath13 (approved):
🧹

PR #337, opened 2025-02-19
refactor: simplify `SortButton` logic

mcmcgrath13:
# PULL REQUEST

## Summary

Remove all of the query selctor/manipulation logic in the sort button. It isn't needed as the sort handler function will trigger updates that will eventually update the props of the sort button so it reflects the correct direction.

This then turned into a bit of a bigger refactor to make sure accessibility was maintained and roles/responsibilities were clear:
1. change approach of `<EcrStyled props...>{table body}</EcrStyled>` to `<EcrTableWrapper><EcrTableHeaders /><EcrTableContent /><EcrTableWrapper>`
  a. This allows a better split on server/client and keeps the headers "the same", so that the focus stays where it was when the sort is toggled
  b. This also lead to reorganizing/renaming of files and matching tests
2. Make the target of the sort the whole table cell area instead of just the icon, but when disabled, only the icon gets greyed out

## Related Issue

Fixes #3200

## Additional Info

Video of interactions with sleep introduced to exaggerate load time:
https://github.com/user-attachments/assets/738d0124-506a-4d59-a392-4be0cb6ba18d


BobanL (approved):
Tested locally and changes look good! Thanks for the refactor!

PR #338, opened 2025-02-19
fix: remove unnecessary key on `Header`

### LINKED ISSUE #25: Fix `Header` key anti-pattern

Fine:
```js
```

Not fine:
```js
```

`Header` currently adds a `key` on the item in isolation of the map. Try to fix this (it might be a deep rabbit hole, so time box).

mcmcgrath13:
# PULL REQUEST

## Summary

Remove the `key` on the jsx returned from `Header`. It's unneccessary.

## Related Issue

Fixes #25


JNygaard-Skylight (approved):
LGTM!

PR #341, opened 2025-02-19
fix: Use dynamic routes to pass eCR ID to view-data page

### LINKED ISSUE #339: Use dynamic routes Instead of an URL parameter for eCR ID
## What needs to be done

Use [NextJS dynamic routes ](https://nextjs.org/docs/14/pages/building-your-application/routing/dynamic-routes)instead of passing the eCR ID as an URL parameter.

## Why it needs to be done

dynamic routes are a more simple and more idiomatic way to get a variable from the URL. It clearly shows that the ID is required.

## Timebox

_How much time should be dedicated to this spike?_

## To-do list

_list out actions to be taken for this spike:_

- [ ]

## Additional context

[OPTIONAL, delete if not needed]

JNygaard-Skylight:
# PULL REQUEST

## Summary

Use [NextJS dynamic routes ](https://nextjs.org/docs/14/pages/building-your-application/routing/dynamic-routes)instead of passing the eCR ID as an URL parameter.

## Related Issue

Fixes #339 

## Additional Infomation
This changes the `view-data` url from: `http://localhost:3000/ecr-viewer/view-data?id=an-ecr-uuid` to `http://localhost:3000/ecr-viewer/view-data/an-ecr-uuid`


mcmcgrath13:
What are the pros/cons of doing it this way vs as a search param?

JNygaard-Skylight:
> What are the pros/cons of doing it this way vs as a search param?

I think the biggest one is that it makes it clear within the code and to the user that the eCR ID cannot be undefined. If you try to go to `view-data/` without an ID, it immediately fails and returns the "Page not found" error page instead of hanging a second or two as we try to retrieve a bundle with `id=undefined` from storage.

mcmcgrath13:
> > What are the pros/cons of doing it this way vs as a search param?
> 
> I think the biggest one is that it makes it clear within the code and to the user that the eCR ID cannot be undefined. If you try to go to `view-data/` without an ID, it immediately fails and returns the "Page not found" error page instead of hanging a second or two as we try to retrieve a bundle with `id=undefined` from storage.

We have a custom not found page for this case vs the general one (which I suppose we could make work here). I'd think the time-to-error page would be about the same (or could be if we did the nullish check on `get_fhir_data`) - are you seeing a big performance difference?

The types of IDs we have look a bit odd to my eye as a url segment, but this is nothing more than gut feel with no backing science or facts

JNygaard-Skylight:
> > > What are the pros/cons of doing it this way vs as a search param?
> > 
> > 
> > I think the biggest one is that it makes it clear within the code and to the user that the eCR ID cannot be undefined. If you try to go to `view-data/` without an ID, it immediately fails and returns the "Page not found" error page instead of hanging a second or two as we try to retrieve a bundle with `id=undefined` from storage.
> 
> We have a custom not found page for this case vs the general one (which I suppose we could make work here). I'd think the time-to-error page would be about the same (or could be if we did the nullish check on `get_fhir_data`) - are you seeing a big performance difference?
> 
> The types of IDs we have look a bit odd to my eye as a url segment, but this is nothing more than gut feel with no backing science or facts

To be honest, I also do not have any evidence, I just think it makes the logic a little nicer, and I actually do prefer the look of the URL. 🤷‍♂️ If you do not think it's worth the hassle I won't be disappointed, just something I wanted to try.
pbdg4l7i1a381 (1) copy

mcmcgrath13:
I don't feel strongly here, but it's a breaking change, so just want to be measured about it (and also if we're going to do it, probably best to do it now and not later) - maybe bring it up as a go back in stand up today to see what more folks think? (or maybe just a slack thread laying out the change)

JNygaard-Skylight:
Closing this because this would require coordination with NBS, with minimal benefit.

PR #350, opened 2025-02-20
ci: create dependabot.yml

BobanL:
# PULL REQUEST

## Summary

- Create dependabot.yml to allow for auto update of our repo

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,12 @@
-# To get started with Dependabot version updates, you'll need to specify which
-# package ecosystems to update and where the package manifests are located.
-# Please see the documentation for all configuration options:
 # https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
 
 version: 2
 updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"
   - package-ecosystem: "npm"
     directory: "/containers/ecr-viewer"
     schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,23 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
```
JNygaard-Skylight, line 7: Should `- package-ecosystem: "github-actions"` be included?

JNygaard-Skylight (changes requested):
Looks good, just curious if we should also consider `github-actions`?

.github/dependabot.yml
```diff
@@ -0,0 +1,25 @@
...
+      interval: "weekly"
+  - package-ecosystem: "npm"
+    directory: "/containers/ecr-viewer"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "docker"
+    directory: "/containers/ecr-viewer"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "pip"
```
mcmcgrath13, line 18: should we also check pip on the fhir converter?

BobanL: Sure! I don't want to add any more though since those others will just get folded into orchestration anyways.

JNygaard-Skylight (approved):
LGTM!

mcmcgrath13 (approved):
🤖

PR #356, opened 2025-02-21
refactor: Fix `PaginationPage` key anti-pattern

### LINKED ISSUE #23: Fix `PaginationPage` key anti-pattern

Fine:
```js
```

Not fine:
```js
```

`PaginationPage` currently adds a `key` on the item in isolation of the map. Try to fix this (it might be a deep rabbit hole, so time box).

jakewheeler:
# PULL REQUEST

## Summary

This PR primarily fixes #23 by removing `key` from the `li` returned by `PaginationPage`. I've also modified the code to make use of the USWDS `Pagination` component props instead of using our own custom type.

## Related Issue

Fixes #23


mcmcgrath13 (approved):
📃

PR #390, opened 2025-02-21
style: enforce plain strings in jsx are not surrounded by curlies

mcmcgrath13:
# PULL REQUEST

## Summary

Enforce via linting that plain strings in jsx are _not_ surrounded by curlies

okay
```
<Component className="my-class"/>
```

not okay
```
<Component className={"my-class"} />
```


angelathe (approved):
🎉

PR #391, opened 2025-02-21
style: enforce explitly true props

mcmcgrath13:
# PULL REQUEST

## Summary

Enforce that properties use explitly true values instead of the implicitly true just key name.

okay
```
<MyComponent prop={true} />
```

not okay
```
<MyComponent prop />
```


BobanL (approved):
✍️

PR #396, opened 2025-02-24
fix: remove unnecessary key on filter condition div

### LINKED ISSUE #24: Fix `FilterReportableConditions` key anti-pattern

Fine:
```js
```

Not fine:
```js
```

`FilterReportableConditions` currently adds a `key` on the item in isolation of the map. Remove it (this one should be trivial).

mcmcgrath13:
# PULL REQUEST

## Summary

Remove the key "all" on a div in filter condition code. It's unnecessary.

## Related Issue

Fixes #24


jakewheeler (approved):
🚢

PR #400, opened 2025-02-24
build: treeshake css

### LINKED ISSUE #389: treeshake bundled css
Our CSS on the `view-data` page (likely the library as well, but I haven't checked) is a whopping 86kb gzipped and about 0.5mb unzipped. This is currently the biggest blocker to our first (loading blob) render.

We currently ship all of the uswds styles when we bundle the app, which in practice we do not use nearly all of. I briefly tried to set up `postcss-purgecss` to fix this, but got weird errors and bailed out.

In practice, browsers are pretty good at caching files like this, so it really hits our fits load per-user-device per-deployment

mcmcgrath13:
# PULL REQUEST

## Summary

Add post processing to make our CSS bundle smaller. This is a pretty conservative pass and takes the gzipped bundled from 84.3kb to 31.7kb.

## Related Issue

Fixes #389

## Additional Information

When using js to construct css classes, we'll need to be sure to add a safelist for those


BobanL (approved):
🤏

PR #403, opened 2025-02-24
refactor: Move database code out of API Directory

### LINKED ISSUE #85: Move database code out of API Directory
## What needs to be done

The code to access Postgres and SQL Server databases are not just used by code within the API directory. Therefore this code should be moved higher in the project directory to better reflect the scope at which they are used. 

## Why it needs to be done

It is confusing that code is unrelated and does not use the APIs would need to import something from within the API directory. 

## Acceptance Criteria

- `postgres_db.ts` and `sqlserver_db.ts`, should have been moved to `containers/ecr-viewer/src/app/services` or `containers/ecr-viewer/src/services`

## Technical Notes

I don't know if it makes the most sense to have a "global" services directory within `containers/ecr-viewer/src` or if `containers/ecr-viewer/src/app/services` works as that. Either one makes more sense than within `containers/ecr-viewer/src/app/api/services`.

jakewheeler:
# PULL REQUEST

## Summary

This PR simply moves the database connection related files out of `/api` and into `/data/db` instead.

## Related Issue

Fixes #85 

## Acceptance Criteria

- [ ] Application should function exactly as it did prior to this change


BobanL (approved):
📦

PR #406, opened 2025-02-25
fix: library user preferences

### LINKED ISSUE #399: Url pararms not respected when local storage set
## Description

When url params are provided, they are ignored if the user storage says otherwise. This leads to confusing behavior.

## Steps to reproduce

1. Navigate to the viewer and go to a page later than one
2. Delete the search params, then go to that page
3. Notice that you get navigated to that later page after a load/flicker

## Expected Result

1. The page number shouldn't be stored in user params at all
2. If the URL specifies a param, it should be respected over the user params

## Actual result

_Explain what's happening now when the bug is present. Attach screenshot if applicable_

## Priority

#### Priority

## Notes

mcmcgrath13:
# PULL REQUEST

## Summary

Fix handling of ecr library user preferences.
1. Remove current page from preferences entirely (would cause weirdness with trying to reset things and then flickering over to a later page)
2. Move the storage to cookies, so it is available on the server side
3. Prefer the search params value to the stored value
4. Make sure the page size is always in the dropdown

## Related Issue

Fixes #399 

## Acceptance Criteria

1. The page number shouldn't be stored in user params at all
2. If the URL specifies a param, it should be respected over the user params

## Additional Information

* This PR adds a bit of a back door where you can set a custom page size using the search params. This could be useful for testing maybe?
* There's a bit of a race-condition-like-thing where if you change the page size, then hit the logo/home button, it won't respect the cookie. This is because of browser caching as far as I can tell (don't even see any network activity). If we decide we care about this, we could add a random cache bust to the link and then add middleware to strip it out. I decided this wasn't worth it, but I could definitely be swayed it is (though probably want to do it as a follow on - we have the same problem with the current `main` branch)


containers/ecr-viewer/src/app/components/LibrarySearch.tsx
```diff
@@ -17,19 +16,8 @@ interface LibrarySearchProps {
  * @returns - Search bar component for the eCR Library
  */
 const LibrarySearch = ({ className, textBoxClassName }: LibrarySearchProps) => {
-  const router = useRouter();
-  const pathname = usePathname();
-  const searchParams = useSearchParams();
+  const { searchParams, updateQueryParam, pushQueryUpdate } = useQueryParam();
```
mcmcgrath13, line 19: this was a side quest refactor when I was trying to figure out some odd caching-related-behavior. But I think it's cleaner, so I left it

BobanL (approved):
Tested locally and it works great!

PR #386, opened 2025-02-21
refactor: reduce duplicated logic

JNygaard-Skylight:
# PULL REQUEST

## Summary

A grab-bag of refactors:
- Use dot notation for FHIR paths: `mappings["patientAddress"]` to `mappings.patientAdrress`.
- Use the specialized `evaluateValue` in a couple of additional places where it makes sense instead of `evaluate`
- Create a function to pull out a human-readable string value for a given `CodeableConcept`. This was done in various ways, with various degrees of rigor.
- Add types to the output of `evaluate` where missing.
- Extract the gender censorship code to its own function.
- Replace `LabReport` type in `labService.tsx` with the correct FHIR type.
-  Combine duplicated code in `ecrSummarySerivce.tsx` for getting relevant resources by SNOMED condition code into a single function.
- Combined duplicated code in `ecrSummaryServive.tsx` and `ecrMetadataService.ts` for getting RCKMS rule summaries   into a single function and put it in a new service `reportabilityService.ts`

## Related Issue

Fixes #84

PR #410, opened 2025-02-25
chore: Enable @typescript-eslint/no-explicit-any ESLint rule 

### LINKED ISSUE #348: Enable `@typescript-eslint/no-explicit-any` ESLint rule
## What needs to be done

This goal of this ticket is to enable the `@typescript-eslint/no-explicit-any` ESLint rule in `containers/ecr-viewer/.eslintrc.json`:
```json
"rules": {
    "@typescript-eslint/no-explicit-any": "error",
}
```

We can consider disabling this rule for test files if we feel it shouldn't apply to non-production code:
```json
"overrides": [
  {
    "files": ["*.test*", "**/tests/**/*"],
    "rules": {
      "@typescript-eslint/no-explicit-any": "off",
      "jsdoc/require-jsdoc": "off"
    }
  }
]
```

With this rule added, running `npm run lint` will provide you with a list of errors that point to all places in the code base where the `any` type was explicitly used. At this point, we will need to:

1. Evaluate each error, determine if `any` is truly needed or if another type is more appropriate
2. Swap out the type, or if `any` makes sense to keep, please add the line `// eslint-disable-next-line @typescript-eslint/no-explicit-any` directly above the code to disable ESLint's check. It may also be helpful for others to add another comment with some additional details around why `any` is needed.

## Why it needs to be done

`any` can have its uses, but it is generally a good idea to avoid it as much as possible since bugs can be introduced when typechecking is disabled. We should do our best to avoid using `any` whenever possible. However, this change will enable us be more intentional about explicitly using `any` if we do need it in some situations.

## Acceptance Criteria

- [ ] ESLint rule has been enabled
- [ ] (optionally) ESLint rule has been disabled for test files
- [ ] Uses of `any` have been inspected and the code has been updated to use a new type, or an exception comment has been added above the offending line
- [ ] No errors occur when running `npm run lint`

## Technical Notes

We are currently using `error: any` in a few spots when catching errors. We may want to consider using `error: unknown` and then asserting the error type so that we can safely use its properties. For example:

```ts
try {
...
} catch (e: unknown) {
    if (e instance of Error) {
        console.log(e.message);
    }
}
```

jakewheeler:
# PULL REQUEST

## Summary

This PR enables the `@typescript-eslint/no-explicit-any` rule and refactors out the usage of `any` where possible. The main objective with this PR is to make it easier on developers to understand the shape of the data they're working with and allow TypeScript to catch potential errors during development.

**Note:** this doesn't mean we can't use `any` at all. You'll notice there are a few instances of `any` that still remain in this PR. This may be due to being particularly tricky to handle without `any` at the moment and/or its usage may not be overly problematic.

## Related Issue

Fixes #348 

## Acceptance Criteria

- [x] Rule is enabled
- [x] Usage of `any` has been eliminated where possible
- [x] Application continues to function as it did previously

## Additional Information

- If there are any remaining `any` types that you think can be eliminated, please let me know!


containers/ecr-viewer/.eslintrc.json
```diff
@@ -43,6 +44,7 @@
     {
       "files": ["*.test*", "**/tests/**/*"],
       "rules": {
+        "@typescript-eslint/no-explicit-any": "off",
```
jakewheeler, line 47: We could leave this rule enabled for test files if folks feel it'd be valuable. I was thinking a little flexibility in non-prod code could be nice to have.

containers/ecr-viewer/src/app/services/labsService.tsx
```diff
@@ -489,7 +496,9 @@ export const evaluateLabInfoData = (
       "Organization/",
       "",
     );
-    const title = report.code.coding.find((c: Coding) => c.display).display;
+    const title = report.code.coding?.find((c: Coding) => c.display)
+      ?.display as string;
```
jakewheeler, line 500: Improving the typing pointed out that `title` could potentially be `undefined`. It would likely be better to handle that situation instead of asserting title as `string` but I wasn't sure exactly what the outcome of that should be.

Any thoughts or suggestions here?

containers/ecr-viewer/src/app/utils/evaluate.ts
```diff
@@ -8,6 +8,8 @@ import {
   UserInvocationTable,
 } from "fhirpath";
 
+// TODO: Follow up on FHIR/fhirpath typing
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jakewheeler, line 12: This seems like a much larger effort that should be left for a future ticket 😓

containers/ecr-viewer/src/app/services/labsService.tsx
JNygaard-Skylight: In my PR #386 I created a new function to handle the logic around getting a human-readable string from a `CodeableConcept`, but it can still return `undefined` (which I think is the right thing to do as what we want as a default is different each time), so I went with this `const title = getHumanReadableCodeableConcept(report.code) ?? "Unknown";`. I don't know if "Unknown" is the best default; that would be a question for @ashton-skylight.

containers/ecr-viewer/src/app/utils/evaluate.ts
JNygaard-Skylight: @mcmcgrath13 just split ticket #416 from her PR #413.

mcmcgrath13: could we use `unknown` instead of `any` - would that get us some better enforcement on downstream usage?

containers/ecr-viewer/src/app/services/labsService.tsx
ashton-skylight: If there's no data, "No data" is acceptable. If the title is undefined, then default to "Unknown" or "Unknown data"

containers/ecr-viewer/src/app/utils/evaluate.ts
jakewheeler: I think `unknown` is too limited and the usage of `evaluate` is too widespread to make this work. In theory, using `unknown` would require us to perform checks on the returned value from `evaluate` (or one of its many wrapper forms) but this feels incredibly difficult to manage as a user of the function, if it is manageable at all. I think we'd end up doing a lot of type assertions to satisfy TS, which doesn't feel much better to me 🫠 

I think taking in `fhirData` as `any` is also okay for now as we're doing proper checks while creating `fhirDataIdentifier`. I noticed our `evaluate` function's signature matches the [fhirpath function's signature](https://github.com/HL7/fhirpath.js/blob/master/src/fhirpath.d.ts#L7), which uses `any`. 

Let me know if I'm off base with what you had in mind @mcmcgrath13. My vision is that we can somehow solve the issue at the `evaluate` level so function callers can receive a real type from it.

containers/ecr-viewer/src/app/services/labsService.tsx
jakewheeler: I've made the change so that in the case of an `undefined` title, we will receive "Unknown data".

containers/ecr-viewer/src/app/services/htmlTableService.ts
```diff
@@ -1,14 +1,15 @@
...
+import React from "react";
 
 interface Metadata {
   [key: string]: string;
 }
 
 export interface HtmlTableJsonRow {
   [key: string]: {
-    value: any;
+    value: string | React.JSX.Element;
```
mcmcgrath13, line 12: maybe use `React.ReactNode` here?
```suggestion
    value: React.ReactNode;
```

mcmcgrath13: or the `RenderableNode` type in data-utils since this probably is a tighter type than `ReactNode`

containers/ecr-viewer/src/app/services/socialHistoryService.tsx
```diff
@@ -22,7 +28,7 @@ export const evaluateTravelHistoryTable = (
     mappings.patientTravelHistory,
   );
 
-  const columnInfo = [
+  const columns: TravelHistoryColumn[] = [
```
mcmcgrath13, line 31:
```suggestion
  const columns: ColumnInfoInput[] = [
```

then delete the `TravelHistoryColumn`

containers/ecr-viewer/src/app/services/htmlTableService.ts
jakewheeler: I think the issue with this was that it needed to be compatible with `RenderableNode`. Although, now that I'm looking at it again, we could do this instead?

```suggestion
    value: RenderableNode;
```

containers/ecr-viewer/src/app/services/socialHistoryService.tsx
```diff
@@ -43,29 +49,49 @@ export const evaluateTravelHistoryTable = (
...
+      });
+
+      return row;
+    })
+    .filter((row) => {
+      // Filter out rows with only noData values
+      return Object.values(row).some((column) => {
+        const columnValue = column.value;
+        return columnValue !== noData;
+      });
```
mcmcgrath13, line 93:
```suggestion
      return Object.values(row).some(({value}) => value !== noData);
```
⛳ (feel free to ignore if you don't like)

containers/ecr-viewer/src/app/services/socialHistoryService.tsx
jakewheeler: Do you think it's worth doing something here to enforce requiring `infoPath`? I think that's what I was mainly after.

containers/ecr-viewer/src/app/view-data/components/EcrDocument/clinical-data.tsx
```diff
@@ -238,27 +247,43 @@ export const returnCareTeamTable = (
     { columnName: "Dates", infoPath: "careTeamParticipantPeriod" },
   ];
 
-  careTeamParticipants.forEach((entry) => {
-    if (entry?.period) {
-      const { start, end } = entry.period;
-      (entry.period as any).text = formatStartEndDate(start, end);
-    }
+  const modifiedCareTeamParticipants: ModifiedCareTeamParticipant[] =
```
mcmcgrath13, line 250: not really for this PR, but this feels like a strange pattern. I wonder if it would be better to modify the `ColumnInfoInput` expectations, so you could provide an `evaluateFn: (entry, mapping) => string` which defaults to `evaluateValue`, but you could override 🤔

containers/ecr-viewer/src/app/view-data/components/EvaluateTable.tsx
```diff
@@ -12,6 +12,9 @@ export interface ColumnInfoInput {
   value?: string;
   className?: string;
   hiddenBaseText?: string;
+  // Generics could be used here instead but we don't want to have to provide it
+  // when this property is optional. Using `any` avoids that issue.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   applyToValue?: (value: any) => any;
```
mcmcgrath13, line 18:
```suggestion
  applyToValue?: (value: string) => string;
```
I'm pretty sure this _should_ be true - does it cause issues?

containers/ecr-viewer/src/app/services/socialHistoryService.tsx
mcmcgrath13: maybe do `Required<Pick<ColumnInfoInput, "fields...">>` to at least make it clear its a subset of that?

containers/ecr-viewer/src/app/view-data/components/EvaluateTable.tsx
jakewheeler: It does. Sometimes we do things like [this](https://github.com/CDCgov/dibbs-ecr-viewer/blob/jw/348/containers/ecr-viewer/src/app/view-data/components/EvaluateTable.tsx#L169) (`data` is a `ReactNode`) 🙈

containers/ecr-viewer/src/app/services/socialHistoryService.tsx
jakewheeler: Yep, I like it. Thanks!

containers/ecr-viewer/src/app/view-data/components/EvaluateTable.tsx
mcmcgrath13: oh `splitStringWith` is the problem. I wonder if we can just move that after the `applyToValue` part. Most of the functions that are passed to `applyToValue` would choke on those `p` tags, so I don't think this is currently functioning if both of those cases hit

jakewheeler: There are some additional issues beyond this example as well, such as [this one](https://github.com/CDCgov/dibbs-ecr-viewer/blob/jw/348/containers/ecr-viewer/src/app/view-data/components/EcrDocument/clinical-data.tsx#L363). `formatDate` optionally accepts a string and returns `string | undefined`.

It looks like `toSentenceCase` used in this same file works similarly to `formateDate`.

containers/ecr-viewer/.eslintrc.json
JNygaard-Skylight: I think I'm in the minority, but I think typing helps write tests as well. You can always use @ts-ignore if you need to.

containers/ecr-viewer/src/app/utils/evaluate.ts
JNygaard-Skylight: To toot my own horn, I think #416 or some other use of generics is the only way to get to any typing with `evaluate`. I think until we tackle that `any` has to do.

containers/ecr-viewer/src/app/view-data/components/common.tsx
```diff
@@ -41,27 +54,42 @@ export const returnImmunizations = (
...
+      };
+
+      newImmunization.occurrenceDateTime = formatDateTime(
+        initialImmunization.occurrenceDateTime ?? "",
+      );
+
+      const manufacturer = evaluateReference<Organization>(
+        fhirBundle,
+        mappings,
+        initialImmunization.manufacturer?.reference || "",
```
JNygaard-Skylight, line 70: `evaluateReference` can now handle `undefined` references.

containers/ecr-viewer/src/app/view-data/components/common.tsx
```diff
@@ -41,27 +54,42 @@ export const returnImmunizations = (
...
-    }
-  });
+  const modifiedImmunizations: ModifiedImmunization[] = immunizationsArray.map(
+    (initialImmunization) => {
+      const newImmunization: ModifiedImmunization = {
+        ...initialImmunization,
+      };
+
+      newImmunization.occurrenceDateTime = formatDateTime(
+        initialImmunization.occurrenceDateTime ?? "",
```
JNygaard-Skylight, line 64: `formatDateTime` can handle `undefined`datetime string.

JNygaard-Skylight (changes requested):

jakewheeler:
Thank you @JNygaard-Skylight! These items should be fixed.

containers/ecr-viewer/src/app/view-data/components/EvaluateTable.tsx
mcmcgrath13: so would `(value: string | undefined) => string | undefined)` cover the usages?

jakewheeler: No 😭  I keep finding new usages each time I make a change lol. [This one](https://github.com/CDCgov/dibbs-ecr-viewer/blob/jw/348/containers/ecr-viewer/src/app/services/labsService.tsx#L397) returns `RenderableNode`.

JNygaard-Skylight (commented):
LGTM to me but I'll eave it to @mcmcgrath13 to approve as they still have an outstanding comment.

mcmcgrath13 (approved):
looks good to me. We can deal with the evaluate table weirdsies as a follow on - would you make a ticket?

jakewheeler:
I've created #428 for the follow up work. Thanks for the review @mcmcgrath13 @JNygaard-Skylight!

JNygaard-Skylight (approved):
Approved!

PR #413, opened 2025-02-25
refactor: bundle fhir paths instead of api return

### LINKED ISSUE #82: SPIKE: Can the use of FHIR paths be simplified?
## What needs to be done

Typically in the eCR Viewer when we want to extract data from the FHIR bundle we use FHIR paths. The FHIR paths are named and defined in `containers/ecr-viewer/src/app/api/fhirPath.yml`. The contents of this YML file is returned along with the requested FHIR bundle through the `save-fhir-data`. The bundle and the mappings are passed throughout the application into virtually all of its different components. 

This process seems reasonable for the FHIR bundle: we need to retrieve an arbitrary bundle from the database, and it needs to be available to be used by the orchestration service. However this seems unnecessarily complex for the FHIR path mappings: The mappings are constant for all bundles, and is not used by any other service.

Instead of passing the FHIR path mappings with the bundle, the mappings could likely be refactored (as an example) as a TypeScript class that can be imported throughout the application.

## Why it needs to be done

The current implementation appears to add complexity for little if any benefit.

## Timebox

_How much time should be dedicated to this spike?_

## To-do list
Determine if there is an acceptable reason for the current implementation
- [ ] Is there a possibility that FHIR path mappings could be unique to a bundle?
- [ ] Is there a possibility that the mappings could be used by other services?
- [ ] Are there any other reasons for the current implementation?
If the answers to the above question are no, then:
- [ ] Determine an alternative implementation that would in fact improve the quality of the codebase, if one is possible.

mcmcgrath13:
# PULL REQUEST

## Summary

What changes are being proposed?

## Related Issue

Fixes #82

## Acceptance Criteria

## Additional Information

## Checklist


containers/ecr-viewer/src/app/tests/components/ActiveProblems.test.tsx
JNygaard-Skylight, line 1: You need to remove `fhirPathMappings` as a parameter to `returnProblemsTable` on line 152. Why is there not an ESLint rule for incorrect number of parameters?

JNygaard-Skylight (changes requested):
LGTM apart from a little issue in the tests.

containers/ecr-viewer/src/app/tests/components/ActiveProblems.test.tsx
mcmcgrath13: looks like tsc catches that, and it's now part of pre-commit, so I think these are all caught now

JNygaard-Skylight (approved):
Approved!

PR #415, opened 2025-02-26
test: fix e2e test flakiness

BobanL:
# PULL REQUEST

## Summary

- Add await on `toHaveURL` assertion
- Add a check to ensure keycloak is up prior to auth tests
- ~~Use `next build` & `next start` during our e2e tests to ensure routes are completely compiled. Previously tests were waiting for routes to compile which could cause them to timeout.~~

### New
- Set ecr-viewer to `network_mode:host` in order to get auth to work in docker-compose.
  -  Without using `network_mode:host` keycloak would need additional configuration to allow communication for both the server (internal docker network) and client (localhost)
  - Orchestration needs to communicate with ecr-viewer via host.docker.internal since it's no longer on the internal docker network.
- Export docker logs after orchestration tests
- Print out response.content when orchestration fails so it's easier to figure out whats wrong rather than just seeing 500 !== 200
- 
## Additional Information

- Ran 10 times in [CI without failure](https://github.com/CDCgov/dibbs-ecr-viewer/actions/runs/13549689133?pr=415)


containers/ecr-viewer/e2e/non-integrated/auth.spec.ts
```diff
@@ -36,7 +36,7 @@ test.describe("keycloak", () => {
         "The eCR Viewer couldn't retrieve the associated eCR file",
       ),
```
mcmcgrath13, line 37: if the non-integrated is still having some flakiness should we add a sleep or higher timeout or something?

mcmcgrath13 (approved):
⏲️

containers/ecr-viewer/playwright.config.ts
```diff
@@ -43,9 +43,10 @@ export default defineConfig({
     },
   ],
   webServer: {
-    command: "npm run local-dev",
+    command: "npm run local-build-start",
```
mcmcgrath13, line 46: why switch to the built vs dev? could we have used `local-docker` instead of adding these new scripts?

containers/ecr-viewer/e2e/non-integrated/auth.spec.ts
```diff
@@ -1,6 +1,24 @@
...
+        const response = await context.get(
+          process.env.AUTH_KEYCLOAK_ISSUER ??
+            "http://localhost:8070/realms/master/",
+        );
+        if (response.status() === 200) {
+          console.log(`Keycloak is ready`);
+          break;
+        }
+      } catch (error) {
+        console.log(`Waiting for Keycloak...`);
```
jakewheeler, line 17: Do you think it'd be useful to handle availability checks during the infrastructure setup so that we can safely assume Keycloak is good to go when our e2e tests start running in the pipeline? I noticed we have an [availability check like this for the app itself](https://github.com/CDCgov/dibbs-ecr-viewer/blob/boban/fix-e2e-auth-tests/.github/workflows/container-ecr-viewer.yaml#L92).

.github/workflows/container-orchestration.yaml
```diff
@@ -132,6 +132,23 @@ jobs:
         working-directory: ./containers/${{env.CONTAINER}}/tests/integration
         run: |
           python -m pytest -m "integration"
+
+      - name: Get docker logs
```
mcmcgrath13, line 136: was this for debugging? or something we want all the time? 30 days retention feels kinda long for this

containers/ecr-viewer/Dockerfile
```diff
@@ -54,16 +60,13 @@ EXPOSE 3000
 ENV APP_VERSION=$APP_VERSION
 ENV PORT 3000
 ENV HOSTNAME "0.0.0.0"
-ENV OTEL_TRACES_EXPORTER=otlp
-ENV OTEL_METRICS_EXPORTER=otlp
+ENV OTEL_TRACES_EXPORTER=none
+ENV OTEL_METRICS_EXPORTER=none
```
mcmcgrath13, line 64: why change these?

containers/ecr-viewer/docker-compose.yaml
```diff
@@ -7,28 +7,37 @@ services:
...
+      - AZURE_STORAGE_CONNECTION_STRING=DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://localhost:10000/devstoreaccount1;
       - AZURE_CONTAINER_NAME=${AZURE_CONTAINER_NAME:-ecr-viewer-files}
       # SQL Server
       - SQL_SERVER_USER=${SQL_SERVER_USER:-sa}
       - SQL_SERVER_PASSWORD=${SQL_SERVER_PASSWORD:-Password1!}
-      - SQL_SERVER_HOST=${SQL_SERVER_HOST:-sqlserver}
+      - SQL_SERVER_HOST=${SQL_SERVER_HOST:-localhost}
       # Orchestration
       - ORCHESTRATION_URL=${ORCHESTRATION_URL:-http://localhost:8080/}
+      # Auth
```
mcmcgrath13, line 31: why so many more env vars now?

containers/ecr-viewer/docker-compose.yaml
```diff
@@ -7,28 +7,37 @@ services:
       dockerfile: Dockerfile
       args:
         - APP_VERSION=1.0
-    ports:
-      - "3000:3000"
+    network_mode: host
```
mcmcgrath13, line 10: how's the performance?

BobanL: It's totally fine! From my testing it was the OTEL communication that was causing the timeout. I'm not sure why the docker network caused that timeout.

containers/ecr-viewer/e2e/non-integrated/auth.spec.ts
BobanL: In the future we may want to move the checks into the job, but right now playwright already handles quite a bit by waiting for `localhost:3000/ecr-viewer` to be available.

mcmcgrath13 (approved):
🚀

PR #423, opened 2025-02-26
style: prefer consts when possible

mcmcgrath13:
# PULL REQUEST

## Summary

When a variable is not re-used, use a `const` instead of `var` or `let`


jakewheeler (approved):
Thanks! 🚢

PR #427, opened 2025-02-27
refactor: replace error throw with console log for missing reportable triggers

JNygaard-Skylight:
# PULL REQUEST

## Summary

If there are no reportable triggers, log an error to the console instead of blowing everything up.

## Related Issue

Fixes # n/a


jakewheeler (approved):
🫡

PR #429, opened 2025-02-27
style: bring order to our imports

mcmcgrath13:
# PULL REQUEST

## Summary

Have consistent styling for imports
```
1. React
2. external packages
3. import from parents (always using "@/app/..." format)
4. import from siblings (always using "./..." format)
5. everything else
```

## Additional Information

Because of the way the aliases work, I changed around how we use pre-commit for linting and type checking by setting up the root as a project and adding scripts and having the necessary dependencies declared there


JNygaard-Skylight (approved):
Beautiful.

PR #463, opened 2025-03-05
feat: Age calculation fixes (app-wide)

### LINKED ISSUE #434: Age calculation fixes (app-wide)
## What needs to be done

- Change all age calculations in the eCR Viewer to use the encounter date, rather than today's date.
- Display ages <2 as month/day instead of year.
- Add units to age values.

## Why it needs to be done

This was another specific request from our TN users. Case investigators and epis looking at eCRs and other medical records want to know the patient's age at the time of the encounter/intervention, not their age as of today's date. 

There's also a usability issue with young patients - currently, their ages display as "0". This is a problem, especially with conditions that receive data from newborn screenings or that focus on cases in children.

Finally, we've heard general feedback from other users that ages without units are confusing.

## Acceptance Criteria

- [ ] "Current age" (in Patient Demographics) is calculated from encounter date when available. When not available, we can use current date as fallback.
- [ ] Patients less than 2 years old display age as a month/day calculation, not year.
    - [ ] Example: Baby born on 1/1/25, and the encounter that generated the eCR happened on 2/5/25. Today their age would display as "0", but should instead format to "1 month, 5 days"
- [ ] All patient ages (in "current age" and "onset age") include a unit . This will be month/days for young patients, or years for patients above 2. (i.e., "10 years" instead of the current "10")
    - [ ] Age at death field should also include units

## Additional context

Work with Ashton to confirm designs for this.

jakewheeler:
# PULL REQUEST

## Summary

This PR modifies displayed ages so that they will not always appear in years. When a patient is under 2 years old, their age will now be represented in a `x months, y days` format. Patents 2 years and older will now be represented as `x years`.

Please see the acceptance criteria for additional details.

Here are a few examples of how you might now see age being represented:
- `26 days`
- `0 days`
- `1 month, 1 day`
- `2 months, 0 days`
- `23 months, 29 days`
- `43 years`

## Related Issue

Fixes #434 

## Acceptance Criteria

- [ ] "Current age" (in Patient Demographics) is calculated from encounter date when available
  - [ ] When encounter date is not available, use today's date
  - [ ] Age in years should have a `years` unit. Ex: `50 years`
  - [ ] If a year is equal to `1`, we should not show the plural form. Ex: `1 year`
- [ ] Patients less than 2 years old have their age represented in `x months, y days`
  - [ ] If a day or month is equal to `1`, we should not show the plural form. Ex: `1 month, 1 day`
- [ ] "Onset age" follows the same rules as "current age"
  - [ ] Use the years `value` of onset age when available
  - [ ] If a years value for onset age is unavailable, fall back to calculating onset age using `onsetDateTime`

## Additional Information

- Does this completely cover all ages in the app? Am I missing anything?


containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
```diff
@@ -155,18 +158,40 @@ export const evaluatePatientDOB = (fhirBundle: Bundle) =>
...
+  if (patientDOBString) {
+    const encounterStartDate = evaluateOne(
+      fhirBundle,
+      fhirPathMappings.encounterStartDate,
+    );
+
+    // use the encounter start date if one is available, otherwise we'll fall back to today's date
+    const laterDate = encounterStartDate
+      ? new Date(encounterStartDate)
+      : new Date();
```
mcmcgrath13, line 189: since `encounterStartDate` will be `undefined` if unavailbe, are both branches of this ternary equivalent?

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
```diff
@@ -155,18 +158,40 @@ export const evaluatePatientDOB = (fhirBundle: Bundle) =>
...
+
+    return getFormattedAge(laterDate, new Date(patientDOBString));
+  }
+
+  return undefined;
```
mcmcgrath13, line 194: the function is way more readable now ❤️

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
```diff
@@ -176,12 +201,14 @@ export const calculatePatientAge = (fhirBundle: Bundle, givenDate?: string) => {
...
   const patientDODString = evaluateOne(fhirBundle, fhirPathMappings.patientDOD);
 
   if (patientDOBString && patientDODString) {
-    const patientDOB = new Date(patientDOBString);
-    const patientDOD = new Date(patientDODString);
-    return dateFns.differenceInYears(patientDOD, patientDOB);
+    const laterDate = new Date(patientDODString);
+    const earlierDate = new Date(patientDOBString);
+
+    return getFormattedAge(laterDate, earlierDate);
```
mcmcgrath13, line 211: in https://github.com/CDCgov/dibbs-ecr-viewer/pull/484 I need the patient age as a number to do math with it - can we separate out the calculating from the formatting in these functions? and/or thoughts on how to best approach this?

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
jakewheeler: They shouldn't be. TS is telling me giving `undefined` to `Date` is invalid. If you do this in JS, you'll receive `Invalid Date`.

mcmcgrath13: interesting, I would have assumed it would have been like passing no arg given how js usually works. wat.

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
jakewheeler: Do you think it'd make the most sense if `calculatePatientAge` created and returned a person's exact `Age` in years, months, and days and then you can decide how to make use of that data? I'm thinking `Age` would look like this:

```ts
interface Age {
  years: number;
  months: number;
  days: number;
}
```
And then some other function (`formatAgeForDisplay`?) would be responsible for taking that and creating/returning the nicely formatted string?

So at the end of the day, if I wanted to get a formatted age to display to a user I might do something like:

```ts
// 2 months, 3 days
 const formattedAge = formatAgeForDisplay(calculatePatientAge(fhirBundle));`.
```
Or in your case:


```ts
const isPatientAnAdult = calculatePatientAge(fhirBundle).years >= 18;
```

What do you think? Any ideas to improve upon this?

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
BobanL: Oh no, it's even worse that that.... 
The [docs say](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#passing_a_non-date_non-string_non-number_value) undefined returns `Invalid Date`, null returns `1970-01-01T00:00:00.000Z`, and if you ever dare to pass in an array its return is based on the implementation in the browser.

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
mcmcgrath13: that sounds good to me! I'd just name the fn `formatAge` and put it in the `formatService.ts`

Another option since the vast majority of the time we want the string back is to have an optional arg to `calculatePatientAge` of `format=true` which does that last step for you, but allows you to get the raw data back if not.  It's a wee bit clunky, but streamlines the common case - not sure if that trade is worth it or not tbh

containers/ecr-viewer/src/app/services/formatService.ts
```diff
@@ -162,3 +167,27 @@ export const formatContactPoint = (
...
+  const { years, months, days } = age;
+
+  if (years >= 2) {
+    return `${years} years`;
+  }
+
+  // when years is under 2, convert them to months and add to total
+  const totalMonths = years * 12 + months;
+
+  return `${totalMonths} month${makePlural(
```
mcmcgrath13, line 190: if the patient is <1 month, do we want to show `0 months, n days` or just `n days`? (cc @ashton-skylight )

ashton-skylight: @mcmcgrath13 Let's only show `n days`

jakewheeler: Made this change and added some tests for it!

containers/ecr-viewer/src/app/tests/services/formatService.test.tsx
```diff
@@ -45,6 +46,55 @@ describe("FormatService tests", () => {
     });
   });
 
+  describe("Format age", () => {
+    it("should only show days", () => {
+      expect(formatAge({ years: 0, months: 0, days: 27 })).toEqual("27 days");
+      expect(formatAge({ years: 0, months: 0, days: 31 })).toEqual("31 days");
+      expect(formatAge({ years: 0, months: 0, days: 1 })).toEqual("1 day");
```
mcmcgrath13, line 53: can we add a `0 days` test just to be sure?

mcmcgrath13 (approved):
👶

containers/ecr-viewer/src/app/tests/services/formatService.test.tsx
jakewheeler: Yes, good call. Added!

PR #466, opened 2025-03-05
chore: rename ECR_DATA table to be lowercase

BobanL:
# PULL REQUEST

## Summary

- ecr_data table should not be BIG.


mcmcgrath13 (approved):
will this require a rebuild?

BobanL:
> will this require a rebuild?

I'm not exactly sure what you mean, but this has no impact on anything. When I was testing locally the case had no impact on any query. The jurisdiction we are testing with did not experience an error either, but did point it out.

mcmcgrath13:
> > will this require a rebuild?
> 
> I'm not exactly sure what you mean, but this has no impact on anything. When I was testing locally the case had no impact on any query. The jurisdiction we are testing with did not experience an error either, but did point it out.

Yup, that's what I was asking. If we developers would need to rebuild our local databases (and therefore real users might need to migrate as well)

PR #471, opened 2025-03-06
fix: update aria-labels to filter buttons when active

### LINKED ISSUE #289: [A11y Library Fix] - Filter boxes should be in an engaged state
## What needs to be done

Add a state marker to the date and condition filters on eCR Library. 

## Why it needs to be done

Because the elements are not currently marked with state (like aria-selected) screenreader users won't know that they've been engaged in the same way that visual users will. The button is marked visually with "last year" so that users know it's active, but this doesn't extend to the aria encoding.

See Axe scan issue: https://cdc-devtools.dequecloud.com/issues/f924a62b-3916-4634-9c60-0930f3eaa26a

## Acceptance Criteria

- [ ] Determine which aria state is most applicable to this filter engaged functionality (checked, pressed, expanded, etc)
- [ ] Apply that aria state to the date and condition filter buttons as applicable

## Technical Details
This is the current state of the date button with the "last year" filter applied:
`<button type="button" class="usa-button margin-right-0 filters-applied" data-testid="button" aria-label="Filter by Received Date" aria-haspopup="listbox" aria-expanded="false">`

We'll need to add an aria element to indicate that the button is engaged. Part of this ticket will be to decide which aria property makes the most sense for this use case - [see the list of supported properties](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes).

austin-hall-skylight:
# PULL REQUEST

## Summary

Updates aria-labels on filter buttons to indicate that they are active. 

## Related Issue

Fixes #289 

## Acceptance Criteria

- [ ] Determine which aria state is most applicable to this filter engaged functionality (checked, pressed, expanded, etc)
- [ ] Apply that aria state to the date and condition filter buttons as applicable

## Additional Information

There was not an aria state that seemed relevant to our need so I used the Google travel site for guidance.


mcmcgrath13 (approved):
❤️

PR #473, opened 2025-03-06
feat!(db): replace all sql parameters will a single sql connection string

BobanL:
*WAITING UNTIL KYSLEY IS MERGED*

# PULL REQUEST

## Summary

- Replace all sql parameters with a single variable, `SQL_SERVER_CONNECTION_STRING`, to allow greater flexibility during setup without introducing additional variables
  - [`ConnectionPool.parseConnectionString`](https://github.com/tediousjs/node-mssql/blob/a2d6d26916006e944fd82023eac9bffca3dbedcf/lib/base/connection-pool.js#L124) converts all available parameters in that connection string into the options object.
- Set revalidate to 10 (seconds) on `/health-check` because the default is [to never refetch the call](https://nextjs.org/docs/14/app/api-reference/file-conventions/route-segment-config#revalidate).
  
## Related Issue

Fixes #

## Acceptance Criteria

## Additional Information

## Checklist


containers/ecr-viewer/.env.sample
```diff
@@ -18,9 +18,7 @@ AWS_SECRET_ACCESS_KEY=xyz
 DATABASE_URL=postgres://postgres:pw@localhost:5432/ecr_viewer_db
 
 # SQL Server
-SQL_SERVER_USER=sa
-SQL_SERVER_PASSWORD=Password1!
-# SQL_SERVER_HOST=sqlserver
+SQL_SERVER_CONNECTION_STRING="Server=localhost;User ID=sa;Password=Password1!;Encrypt=false;"
```
mcmcgrath13, line 21: why not just use `DATABASE_URL` for both of these and then know which to use based on the config?

BobanL: That would definitely simplify this!

BobanL:
> how does this play with the new kysely stuff?

Unfortunately i don't think it plays nicely at all since this uses specific features in the mssql library. We could look importing the library just to use that function, OR we can copy that logic. https://github.com/tediousjs/node-mssql/blob/a2d6d26916006e944fd82023eac9bffca3dbedcf/lib/base/connection-pool.js#L124

mcmcgrath13:
> > how does this play with the new kysely stuff?
> 
> Unfortunately i don't think it plays nicely at all since this uses specific features in the mssql library. We could look importing the library just to use that function, OR we can copy that logic. https://github.com/tediousjs/node-mssql/blob/a2d6d26916006e944fd82023eac9bffca3dbedcf/lib/base/connection-pool.js#L124

@akasper what do you think?

akasper:
Kysely VASTLY prefers using a connection configuration object, and if we ever add support for more database dialects, I think we will prefer that as well. I'm open to modifying so that we only ever use a connection string if that's what the team prefers.

PR #482, opened 2025-03-07
fix: change ecr_viewer_url to use host.docker.internal

BobanL:
# PULL REQUEST

## Summary

- Update ecr_viewer_url to host.docker.internal since ecr-viewer now runs on host network mode


mcmcgrath13 (approved):
🤖

PR #486, opened 2025-03-07
feat: add search param validation middleware

### LINKED ISSUE #430: URL syncing & validation to filter/search parameters in eCR Library
## What needs to be done
**Syncing:**
URL params should always be synced with the filter/search term(s) selected in the UI -- if invalid inputs are given, URL params should reset to a default state with the UI. 

**Validation:**
 Move URL validation logic to middleware level to ensure that all filter and search parameters (ex. `dateRange`, `dates`, `condition`, `search`) receive valid inputs. Validation logic should already exist for `dates` and `dateRange`, but might need to add logic for `condition` param.

The validation/syncing should handle:
- incorrect formats
- unexpected characters, typos
- malicious inputs (?)

## Current behavior
GIVEN the eCR Library
WHEN a user adds bad inputs to the URL parameters (i.e., dateRange that is not one of the predefined choices, invalid custom dates, invalid condition option(s))
THEN
- Date filter resets to default, BUT `dateRange` and/or `dates` params still display bad inputs.
- Conditions filter resets to displaying 0 conditions, BUT 'condition' param still displays bad inputs.

## Why it needs to be done
Currently, if bad inputs are manually added to the URL params, the URL params and the UI fall out of sync -- this is very confusing / bad UI. Also, bad URL parameters can cause errors leading to incorrect filtering/search results. Adding validation/syncing will improve data integrity and improve the user experience

## Acceptance Criteria
- [ ] All filter and search parameters (dateRange, dates, condition, etc.) are validated before processing.
    - [ ] Should the validation logic be moved/handled at the middleware level?
    - [ ] Invalid inputs will reset filters to their default state. (this should already be done for at least 'dates' and 'dateRange`)
- [ ] If invalid inputs are manually added to the URL params, the URL param(s) in question should automatically reset to their default state, ensuring the URL remains in sync with the filter logic.
- Validation and syncing logic should be moved to middleware level.
- Tests cover valid and invalid cases for all relevant parameters (incorrect formats, invalid chars, etc.).

## Relevant Files
- `src/app/components/Filters.tsx`
- `src/app/components/BaseFilter.tsx`
- `src/app/utils/date-utils.ts` - `isValidParamDates`, `returnParamDates`, etc.

## Additional info
- I can't access the relevant issues/PRs that worked on search/filters, but message me, I might have old notes on this work for context

mcmcgrath13:
# PULL REQUEST

## Summary

Add search param validation middleware and generally overhaul library search param usage so that it is consolidated and typed.

* add `withUrlParamChecks` middleware which makes sure all params we expect exist only once and are valid, typically by deleting invalid params and then the default will get set appropriately later
  * We don't currently delete params we don't expect - should we? could maybe play weird with auth or similar
* Create a `search-param-utils.ts` file which defines a specification for the library search params (default and validator) and generates a config based on the search params and cookies. This is then used in the home page component in lieu of individually getting all the params
* In components that were previously querying the params, pass them down via props instead
  * This allows some simplification around effects reacting to search param changes
  * Generally, components should only update the params, but get needed data from their parents

## Related Issue

Fixes #430

## Acceptance Criteria

- [ ] All filter and search parameters (dateRange, dates, condition, etc.) are validated before processing.
    - [x] Should the validation logic be moved/handled at the middleware level?
    - [ ] Invalid inputs will reset filters to their default state. (this should already be done for at least 'dates' and 'dateRange`)
- [ ] If invalid inputs are manually added to the URL params, the URL param(s) in question should automatically reset to their default state, ensuring the URL remains in sync with the filter logic.
- Validation and syncing logic should be moved to middleware level.
- Tests cover valid and invalid cases for all relevant parameters (incorrect formats, invalid chars, etc.).


containers/ecr-viewer/src/middlewares/withUrlParamChecks.ts
```diff
@@ -0,0 +1,84 @@
...
+    const param = searchParams.get("columnId") as string;
+    const validIds = INITIAL_HEADERS.filter((h) => h.dataSortable).map(
+      (h) => h.id,
+    );
+    if (!validIds.includes(param)) {
+      // if we're deleteing the column, doesn't make sense to keep the direciton
+      return ["columnId", "direction"];
+    }
+  },
+  page: isPositiveInt("page"),
```
mcmcgrath13, line 42: should this check the page is in range? or is that too heavy for middleware?

angelathe:
Re: not deleting params we don't expect, I wonder if we want to consider deleting all params (ones we expect & don't expect) when clicking the reset button? Or would this still have weird effects on auth?

This came up when I tried injecting `&asdf=asdf` after some other params, and clicking the reset button didn't remove that junk param

mcmcgrath13:
> Re: not deleting params we don't expect, I wonder if we want to consider deleting all params (ones we expect & don't expect) when clicking the reset button? Or would this still have weird effects on auth?
> 
> This came up when I tried injecting `&asdf=asdf` after some other params, and clicking the reset button didn't remove that junk param

I've also wondered if the `reset` button should delete everything and/or if we should filter out unexpected params, but the downside seems higher than the upside for me as the extra params do no harm, just junk up the url. The downside being we accidentally delete something that was providing a function for auth or i18n or similar and now we broke something with a hard to debug interaction effect

The main pro I see is someone had a typo and we'd remove it and that might give a hint, but I'm not sure that's worth it for the type of users we expect

What do you see as the trade offs?

containers/ecr-viewer/src/middlewares/withUrlParamChecks.ts
```diff
@@ -0,0 +1,42 @@
+import { NextRequest, NextResponse } from "next/server";
+
+import { LIBRARY_SEARCH_PARAMS } from "@/app/utils/search-param-utils";
+import { ChainableMiddleware, MiddlewareFactory } from "@/middleware";
+
+// This is currently hard coded on the library search params, but could be made configurable
+// with a matcher down the road.
+
+/**
+ * Checks that all URL params are valid and deletes them and redirects the url if not.
```
angelathe, line 10: I would write a note here that says we don't delete params we don't expect.

containers/ecr-viewer/src/app/tests/utils/search-param-utils.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+    it("rejects nonsense headers for dates", () => {
+      const params = new URLSearchParams("dates=123");
+      validator!(params);
+      expect(params.toString()).toBe("");
+    });
+    it("rejects headers for dates without range", () => {
+      const params = new URLSearchParams("dates=2023-01-01|2023-01-01");
+      validator!(params);
+      expect(params.toString()).toBe("");
+    });
```
angelathe, line 113: Maybe I'm misunderstanding the test, but wouldn't having the same start/end date still be valid? Custom date searches from midnight on start date -> 11:59PM on end date.

containers/ecr-viewer/src/app/utils/date-utils.ts
```diff
@@ -138,8 +140,8 @@ const DATE_PARAM_REGEX = /^\d{4}-\d{2}-\d{2}\|\d{4}-\d{2}-\d{2}$/;
  * @returns A boolean; true if the params are valid, false if invalid.
  */
 export function isValidParamDates(
```
angelathe, line 142: This actually may be overkill, but do we want to check whether or not start date < end date here? I know the custom dates UI automatically checks it, but if someone manually changes the dateRange param to, the query will still go through

containers/ecr-viewer/src/app/utils/search-param-utils.ts
```diff
@@ -0,0 +1,172 @@
...
+/**
+ * A function, which given the search params, makes the search params valid, typically
+ * by deleting offending params
+ */
+type ValidationFn = (searchParams: URLSearchParams) => void;
+
+export type PageSearchParams = { [key: string]: string | string[] | undefined };
+
+// helper to check date paramaters are valid, used for both params
+const checkDates: ValidationFn = (searchParams: URLSearchParams) => {
```
angelathe, line 21: Thoughts on moving this to `date-utils` and just importing `checkDates` instead of `isValidParamDates`?

angelathe (commented):
Mainly had questions, but thank you for doing this!!

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -275,43 +260,36 @@ const FilterReportableConditions = ({ conditions }: FilterProps) => {
...
-    } else {
-      setFilterDateOption(DEFAULT_DATE_RANGE);
-      setStartDate("");
-      setEndDate("");
-    }
-  };
-
-  // on mount, make sure the filters match the search params
-  useEffect(resetFilterDate, []);
+  // Keep state in sync with updated params while maintaining correct focus on submit
```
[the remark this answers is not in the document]
mcmcgrath13: let me check if that part is still real or I imagined it mid-debugging something and it was a red herring

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -275,43 +260,36 @@ const FilterReportableConditions = ({ conditions }: FilterProps) => {
...
+  useEffect(() => {
+    setStartDate(initStart);
+    setEndDate(initEnd);
+  }, [initCustomDate]);
```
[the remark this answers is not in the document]
mcmcgrath13: line 276 will always run on every render of the component (so when a prop or state changes)

containers/ecr-viewer/src/app/utils/date-utils.ts
```diff
@@ -138,8 +140,8 @@ const DATE_PARAM_REGEX = /^\d{4}-\d{2}-\d{2}\|\d{4}-\d{2}-\d{2}$/;
...
-  dateRange: string,
-  datesParam: string | undefined,
+  dateRange: string | null,
+  datesParam: string | null | undefined,
```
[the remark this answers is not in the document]
mcmcgrath13: yeah, it's a bit ugly, but basically when we don't get a param back from the query string we get a `null`, so we want to handle that. Maybe we don't need the undefined anymore 🤔

containers/ecr-viewer/src/app/tests/utils/search-param-utils.test.tsx
mcmcgrath13: the issue is we don't have a `dateRange`, not the dates themselves, those are valid

containers/ecr-viewer/src/app/utils/search-param-utils.ts
mcmcgrath13: we use them in two different calling contexts, so I don't think it makes sense to consolidate. The search params come in different formats depending on whether its middleware or page, so we need that part to be handled outside

angelathe:
> > Re: not deleting params we don't expect, I wonder if we want to consider deleting all params (ones we expect & don't expect) when clicking the reset button? Or would this still have weird effects on auth?
> > This came up when I tried injecting `&asdf=asdf` after some other params, and clicking the reset button didn't remove that junk param
> 
> I've also wondered if the `reset` button should delete everything and/or if we should filter out unexpected params, but the downside seems higher than the upside for me as the extra params do no harm, just junk up the url. The downside being we accidentally delete something that was providing a function for auth or i18n or similar and now we broke something with a hard to debug interaction effect
> 
> The main pro I see is someone had a typo and we'd remove it and that might give a hint, but I'm not sure that's worth it for the type of users we expect
> 
> What do you see as the trade offs?

Hm, what would the auth or i18n-related params be ahead of time? / Would they be static where we could have the resetHandler remove all but those params from the URL?

But I think you're right, I think my main pro was just maintaining a clean URL/fixing typos for users. Not the most pressing issue probably.

containers/ecr-viewer/src/app/components/Filters.tsx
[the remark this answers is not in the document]
mcmcgrath13: okay, yup, it's real. Basically the alternate solution here was a key on the component, but that would have broken the connection on the ref so when you hit apply, the focus didn't go back to the main button

containers/ecr-viewer/src/app/utils/date-utils.ts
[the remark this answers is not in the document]
mcmcgrath13: was able to get rid of the undefined, so now it's just sometimes null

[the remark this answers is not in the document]
mcmcgrath13: fixed and docstring updated

containers/ecr-viewer/src/app/tests/utils/search-param-utils.test.tsx
angelathe: Ah got it got it, thanks

angelathe (approved):
🎉🎉🎉

PR #511, opened 2025-03-14
feat!: simplify auth vars

BobanL:
# PULL REQUEST

## Summary

- Replace idp specific variables with generic AUTH ones
- Add AUTH_PROVIDER so we know which one is being used.


containers/ecr-viewer/.env.sample
```diff
@@ -28,12 +28,11 @@ ORCHESTRATION_URL=http://localhost:8080
...
-AUTH_KEYCLOAK_SECRET=P7e6LOwZXhJq1MCo0GSRAQFqBOYDuaBx
-AUTH_KEYCLOAK_ISSUER=http://localhost:8070/realms/master
-# AUTH_AZURE_AD_ID=
-# AUTH_AZURE_AD_SECRET=
-# AUTH_AZURE_AD_TENANT_ID=
+AUTH_PROVIDER=keycloak
+AUTH_CLIENT_ID=ecr-viewer
+AUTH_CLIENT_SECRET=P7e6LOwZXhJq1MCo0GSRAQFqBOYDuaBx
+AUTH_ISSUER=http://localhost:8070/realms/master
+# AUTH_TENANT_ID=
```
mcmcgrath13, line 35: is an issuer equivalent to a tenant?

BobanL: No, azure ad in nextauth requires tenant_id and keycloak requires issuer. I think azure ad is the only one that requires `tenant_id`, but there are a few id providers that require issuer. 


NextAuth v5 (still in beta) uses issuer for Entra (azure ad) as well. https://authjs.dev/getting-started/providers/microsoft-entra-id

BobanL: To simplify things further we could pretend that you could use `AUTH_ISSUER` in the place of `AUTH_TENANT_ID` even though one is usually a url and the other is a uuid.

mcmcgrath13: do you think it's nicer to have the one env var with some overhead on documentation if using azure ad? or have the specific env var, but then more env vars to manage?

BobanL: I think it makes more sense to squash these down into a single AUTH_ISSUER variable with additional documentation (currently in the terraform).

mcmcgrath13 (approved):
🤏

PR #532, opened 2025-03-17
feat: replace eRSD & RCKMS data with TES API data in TCRS

### LINKED ISSUE #54: Use TES API in TCRS
## What needs to be done

Replace the various scripts that retrieve data for the TCRS with a single script that uses the TES API.

## Why it needs to be done

The TCRS currently uses multiple scripts that use multiple APIs to acquire information on conditions and their related concepts. Most if not all of it can be replaced with a script that uses the TES API. [See the previous SPIKE](https://app.zenhub.com/workspaces/streamline-ecr-6480bf2ee530095ab41ebbe9/issues/gh/cdcgov/phdi/2915). 

## Acceptance Criteria

- TCRS is using the TES as the source of truth
- There should be only one script to acquire all of the data the TCRS needs.
- TCRS should use only one database to store all its data.
- No other CSVs, zips, etc, should be needed
- All old DBs, scripts, CSVs, etc, are removed.

## Technical Notes

There is a branch on our old phdi repo that has most of this already implemented.

jakewheeler:
# PULL REQUEST

## Summary

This PR modifies the `trigger-code-reference` service to remove the dependencies on eRSD and RCKMS data and replace it with data pulled in from the [Terminology Exchange Service](https://tes.tools.aimsplatform.org/) (TES) API. 

A script is used to pull condition and concept data from the TES API and create/load the `data/tes.db` SQLite database, which is queried by the TCRS.

## Related Issue

Fixes #54 

## How to test

There are two main pieces to this work that can be manually tested:

1. Running the seed script and creating the SQLite DB containing the TES data
2. Running the TCRS and testing some data against the endpoints

### Creating/seeding the DB

I have included a `tes.db` file from a recent run of the seed script as part of this PR, but if you'd like to try creating it yourself please take a look at the newly created `data/README.md` file. This should guide you through the process.

### Manual TCRS checking

If you'd like to try running the TCRS locally and testing the endpoints you can follow the existing documentation [here](https://github.com/CDCgov/dibbs-ecr-viewer/blob/main/containers/trigger-code-reference/README.md).

The results returned from the API will be similar to what we saw prior to this change. However, you may notice some additional concept codes coming back from the `/get-value-sets` endpoint. TES offers ~10k additional concepts compared to eRSD, so the data we receive back from this endpoint will be a bit more thorough.

## Acceptance Criteria

- [x] TCRS is using the TES as the source of truth
- [x] There should be only one script to acquire all of the data the TCRS needs
- [x] TCRS should use only one database to store all its data
- [x] All old DBs, scripts, CSVs, etc, are removed*

*This is the case except for the files needed to build the ICD10 -> ICD9 crosswalk table. TES did not return ICD9 codes.

## Additional Information

-


containers/trigger-code-reference/data/README.md
```diff
@@ -0,0 +1,16 @@
...
+Running the `seed_tes_data.py` script will create and populate the `tes.db` SQLite database with data. This database will then be queried to fetch condition and concept information used by the TCRS.
+
+## Prerequisites
+
+Before you can run the seed script, you'll need to acquire a TES API key. Make a TES account [here](https://tes.tools.aimsplatform.org/) and the `API-KEY` menu will be available to you once you log in.
+
+## Running the script
+
+1. Get set up to run the TCRS locally ([see here](../README.md#running-from-python-source-code))
+2. Ensure dependencies have been installed
```
BobanL, line 12: Can we add a note about installing both `requirements.txt` and `dev-requirements.txt`?

BobanL (approved):
From my perspective it looks good! Using the limited data we have for testing the results were exactly the same 🎉

containers/trigger-code-reference/data/README.md
jakewheeler: Yes, definitely!

PR #539, opened 2025-03-18
fix: add revalidate on /health-check to force refresh

BobanL:
# PULL REQUEST

## Summary

- [Add `revalidate` variable](https://nextjs.org/docs/14/app/api-reference/file-conventions/route-segment-config#revalidate) to ensure /health-check will be re-rendered 

## Additional Information

- The /health-check endpoint will never update the values after build time without this change. This issue does not occur when running in dev mode.


mcmcgrath13 (approved):
nice catch!

PR #546, opened 2025-03-19
feat: Add Kysely-driven migrations

### LINKED ISSUE #29: Implement automated database migrations
## What needs to be done

Use Kysely's migrations to automate database setup and updates.

## Why it needs to be done

This will reduce the initial setup burden on our partners, as well as allowing us to make changes to the metadata schemas without needing them to run sql scripts.

## Acceptance Criteria

- Create migration scripts for existing SQL scripts
- On application startup, run any migrations that haven't already been run

akasper:
# PULL REQUEST

_PR title: Remember to name your PR descriptively and follow [conventional commits](https://cheatsheets.zip/conventional-commits)!_

## Summary

* Adds API route and service to explicitly execute migrations
  * API route protected by secret that deployer sets or is randomly generated and logged to server
* Removes `./sql/core.sql` and `./sql/extended.sql`-driven schema initialization
* Adds initial migration to `common` schema
* Adds initial migration to `core` schema
* Adds initial migration to `extended` schema
* Adds various Kysely-driven schema introspection utilities
* Add error UI for incomplete db setup

## Related Issue

Fixes #29 

## Acceptance Criteria

- [ ] Create migration scripts for existing SQL scripts
- [ ] On application startup, run any migrations that haven't already been run
  - Instead of running on startup, we log on startup if there are pending migrations, admin must run migration manually via API route

## Additional Information

* The initial migration scripts (`common/19700101000000_initial.ts`, `core/19700101000000_initial.ts`, and `extended/197001010000000_initial.ts`) check for the presence of an existing schema. This is to ensure that the migrations are not destructive in an already-productionalized environment.


containers/ecr-viewer/src/app/data/db/schemas/core/19700101000000_initial.ts
```diff
@@ -0,0 +1,92 @@
+import { Kysely, sql } from "kysely";
+
+/**
+ * Based on ecr-viewer/sql/core.sql.
+ * Core schema initialization. This version is hard-coded to the
+ * original postgres implementation. Future versions will be
+ * database-agnostic.
```
mcmcgrath13, line 7: my understanding of migrations is that they need to build on each other, so this will always be applied as the base and then the next one assumes this one is already applied, so shouldn't this also be DB agnostic (or at least play nice with both?)

containers/ecr-viewer/src/app/data/db/schemas/core/19700101000000_initial.ts
```diff
@@ -0,0 +1,92 @@
...
+  // Check if schema exists in Postgres
+  const schemaExists = await db
+    .selectFrom("information_schema.schemata")
+    .select("schema_name")
+    .where("schema_name", "=", "ecr_viewer")
+    .executeTakeFirst();
+
+  if (schemaExists) {
+    console.log(
+      "Schema ecr_viewer already exists in core database. Skipping table creation.",
```
mcmcgrath13, line 22: can we assume if the schema exists, that everything is fine?

akasper: I think so, at least for current live deployments.

containers/ecr-viewer/src/app/data/db/schemas/core/19700101000000_initial.ts
akasper: Fair enough. My take is that the initial implementation can be hard-coded to the currently-deployed environments, then we can retroactively genericize these when we have a better type mapping system. 

But I see your point. I'll switch these to use `getSql` or something similar now instead.

containers/ecr-viewer/src/app/data/db/utils/migrate.ts
```diff
@@ -0,0 +1,178 @@
...
+  migrationsDir: string,
+  command: string,
+  target?: string,
+) {
+  const migrator = new Migrator({
+    db,
+    provider: new FileMigrationProvider({
+      fs,
+      path,
+      migrationFolder: migrationsDir,
```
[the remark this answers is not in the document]
akasper: Fixing this right now. I’m going to provide a path that supports multiple migrations directories. If this isn’t doable, then I’m instead going to move that we (for now) just stick to the original (`core|extended` without `common`) directory approach.

containers/ecr-viewer/src/instrumentation.test.ts
```diff
@@ -22,11 +22,13 @@ describe("register and and setupConfigurationVariables", () => {
 
   it("should set AWS_INTEGRATED configuration variables", async () => {
     process.env.CONFIG_NAME = "AWS_INTEGRATED";
+    const jestSpy = jest.spyOn(console, "warn").mockImplementation();
     await register();
 
     expect(process.env.NBS_AUTH).toBe("true");
     expect(process.env.NON_INTEGRATED_VIEWER).toBe("false");
     expect(process.env.SOURCE).toBe("s3");
+    expect(jestSpy).toHaveBeenCalled();
```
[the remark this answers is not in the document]
akasper: Because I want to explicitly warn on startup that if you are running in an environment that doesn’t have a database, that migrations will not be run. The alternative is to potentially silently “fail” to run migrations when a dev or admin is expecting them to be run.

containers/ecr-viewer/src/instrumentation.ts
```diff
@@ -56,3 +57,12 @@ function setupConfigurationVariables() {
   }
   makeEnvPublic(["NON_INTEGRATED_VIEWER"]);
 }
+
+/**
+ * Run database migrations by importing migrate package with side effects
+ */
+async function runDatabaseMigrations() {
```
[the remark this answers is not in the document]
akasper: I think this should be a follow-on ticket. As implemented, using the legacy `.sql` scripts allows the existing containers to keep working as-is. This reduces the potential risks associated with the migrations rollout, IMO, and breaks container maintenance into its own ticket. I welcome your thoughts @mcmcgrath13

akasper:
The biggest blocker on this right now is understanding how best to automatically run the migrations. 

On the one hand, @alismx suggested that this ought not to be done with commands, and perhaps instead should be triggered as part of app launch. Part of the reason for this is that we don't want to meaningfully modify the Dockerfile in order to support the addition of this behavioral change in every environment. (I'm willing to be corrected on my understanding here.) And perhaps that's for the better; running migrations directly via a command would (in the simplest case) require the addition of the `ts-node` package. 

But then comes the question: Where in the startup stack do we want to inject running migrations? Should we do it the same way in both our integration testing environments and our productionalized runtime environments? 

The internet -- bless it -- seems to think the most straightforward way of automatically running migrations on bootup of a NextJS app is to call a `migrate()` function (in our case `migrateUp()`) during the call to `register()` in `instrumentation.ts`. But... does that even work? What about our test environments?

Another suggestion was that we could modify our `build` command to use some sort of specialized config that runs migrations in a separate file... something like:

```json
// tsconfig.server.json
{
  "extends": "./tsconfig.json",
  "compilerOptions": {
    "outDir": "./dist",
    "module": "commonjs",
    "target": "es6",
    "noEmit": false
  },
  "include": ["run-migrations.ts", "jest-global-setup.ts"]
}
```

...and...

```
// package.json
{
  "scripts": {
    "build": "next build && tsc --project tsconfig.server.json"
  }
}
```

Is this approach kosher? If so, I would also have to modify jest config to include a `globalSetup`:

```javascript
module.exports = {
  // ... other Jest stuff
  globalSetup: '<the normal pathing>/dist/jest-global-setup.js'
};
```

...and...

```ts
// jest-global-setup.ts
import { runMigrations } from './run-migrations'; // Import the migration runner

export default async function globalSetup() {
  await runMigrations();
}
```

No idea if this will work. Looking for guidance from someone with more TypeScript/Docker/Jest experience than me. @austin-hall-skylight @JNygaard-Skylight or @mcmcgrath13?

containers/ecr-viewer/src/app/data/db/utils/migrate.ts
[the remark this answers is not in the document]
akasper: I've added a `MultiDirectoryMigrationProvider` that combines the migrations from -- get this -- multiple directories in order to provide migrations to the migration runner.

containers/ecr-viewer/integration/data/db/schema/core.test.ts
```diff
@@ -0,0 +1,479 @@
...
+
+interface MigrationRow {
+  name: string;
+  timestamp?: string;
+}
+
+describe("Common Schema: ", () => {
+  /** 
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+
```
mcmcgrath13, line 18: delete commented out code?

containers/ecr-viewer/integration/data/db/schema/extended.test.ts
```diff
@@ -0,0 +1,1848 @@
...
+  name: string;
+  timestamp?: string;
+}
+
+describe("Extended Schema: ", () => {
+  /** 
+CREATE SCHEMA ecr_viewer
+
+CREATE TABLE ecr_viewer.ecr_data
+(
```
mcmcgrath13, line 20: delete?

containers/ecr-viewer/src/app/data/db/dialects/postgres/utils.ts
```diff
@@ -0,0 +1,148 @@
...
+  tableName: string,
+  columnName: string,
+): Promise<boolean> {
+  const result = await kysely
+    .selectFrom("information_schema.columns")
+    .select("column_name")
+    .where("table_schema", "=", schemaName)
+    .where("table_name", "=", tableName)
+    .where("column_name", "=", columnName)
+    .executeTakeFirst();
```
mcmcgrath13, line 67:
```suggestion
  const result = await getColumn(kysely, schemaName, tableName, columnName);
```

containers/ecr-viewer/src/app/data/db/schemas/common/19700101000000_initial.ts
```diff
@@ -0,0 +1,86 @@
...
+      .executeTakeFirst();
+  } else {
+    result = await db
+      .selectFrom("information_schema.schemata")
+      .select("schema_name")
+      .where("schema_name", "=", schema)
+      .executeTakeFirst();
+  }
+
+  console.log("Result: " + result);
```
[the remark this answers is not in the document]
mcmcgrath13: result doesn't seem to be used besides logging - can all this be removed?

containers/ecr-viewer/src/app/data/db/schemas/common/19700101000000_initial.ts
```diff
@@ -0,0 +1,86 @@
+// import { Kysely } from "kysely";
+import { Kysely } from "kysely";
+
+import { dbNamespace } from "@/app/api/services/database";
+import { getSql } from "@/app/api/services/dialects/common";
+
+const schema = dbNamespace();
```
mcmcgrath13, line 7: I think we probably need to move this off of the top level to make sure it's getting the current environment variables given this gets set up via instrumentation - have you run into any issues with that?

containers/ecr-viewer/src/app/data/db/schemas/core/19700101000001_initial.ts
```diff
@@ -0,0 +1,62 @@
+import { Kysely, sql } from "kysely";
+
+import { dbSchema } from "@/app/api/services/database";
+import { getSql } from "@/app/api/services/dialects/common";
+
+/**
+ * Based on ecr-viewer/sql/core.sql.
+ * Core schema initialization. This version is hard-coded to the
+ * original postgres implementation. Future versions will be
+ * database-agnostic.
```
mcmcgrath13, line 10: is this still true?

containers/ecr-viewer/src/app/data/db/schemas/core/19700101000001_initial.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export async function up(db: Kysely<any>): Promise<void> {
+  if (dbSchema() === "extended") {
+    console.log("Extended schema detected. Skipping core migration.");
+    return;
+  }
+
+  // Install uuid-ossp extension (Postgres-specific)
+  await sql`CREATE EXTENSION IF NOT EXISTS "uuid-ossp";`.execute(db);
```
mcmcgrath13, line 21: can we guard this in a postgres check?

mcmcgrath13: also, should this move after the `coreCheck`?

containers/ecr-viewer/src/app/data/db/schemas/core/19700101000001_initial.ts
```diff
@@ -0,0 +1,62 @@
...
+  await db.schema
+    .dropTable("ecr_viewer.ecr_rr_rule_summaries")
+    .ifExists()
+    .execute();
+  await db.schema
+    .dropTable("ecr_viewer.ecr_rr_conditions")
+    .ifExists()
+    .execute();
+  await db.schema.dropTable("ecr_viewer.ecr_data").ifExists().execute();
+  await db.schema.dropSchema("ecr_viewer").ifExists().execute();
```
mcmcgrath13, line 61: strictly speaking, shouldn't this remove the columns added here? or at least only drop the ecr_data?

containers/ecr-viewer/src/app/data/db/schemas/extended/19700101000001_initial.ts
```diff
@@ -0,0 +1,132 @@
+import { Kysely } from "kysely";
+
+import { dbSchema } from "@/app/api/services/database";
+import { getSql } from "@/app/api/services/dialects/common";
+
+/**
+ * Based on ecr-viewer/sql/extended.sql.
+ * Extended schema initialization. This version is hard-coded to the
+ * original SQL Server implementation. Future versions will be
+ * database-agnostic.
```
mcmcgrath13, line 10: still true?

containers/ecr-viewer/src/app/data/db/schemas/extended/19700101000001_initial.ts
```diff
@@ -0,0 +1,132 @@
...
+  // Kysely requires <any>
+
+  if (dbSchema() === "core") {
+    console.log("Core schema detected. Skipping extended migration.");
+    return;
+  }
+
+  const extendedCheck = await db
+    .selectFrom("ecr_viewer.ecr_data")
+    .select("patient_name_first")
```
mcmcgrath13, line 24: `patient_name_first` never exists in extended - should this be `first_name`?

containers/ecr-viewer/src/app/data/db/schemas/extended/19700101000001_initial.ts
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Based on ecr-viewer/sql/extended.sql.
+ * Extended schema initialization. This version is hard-coded to the
+ * original SQL Server implementation. Future versions will be
+ * database-agnostic.
+ * @param db - the database connection
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export async function up(db: Kysely<any>): Promise<void> {
+  // Kysely requires <any>
```
mcmcgrath13, line 15: is this true here? or could this be a `Kysely<Common>`? Or because migrations are kind of operating outside of the normal world, this is the way to go?

containers/ecr-viewer/src/app/data/db/schemas/extended/19700101000001_initial.ts
```diff
@@ -0,0 +1,132 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export async function down(db: Kysely<any>): Promise<void> {
+  await db.schema.dropTable("ecr_viewer.ecr_labs").ifExists().execute();
+  await db.schema
+    .dropTable("ecr_viewer.ecr_rr_rule_summaries")
+    .ifExists()
+    .execute();
+  await db.schema
+    .dropTable("ecr_viewer.ecr_rr_conditions")
+    .ifExists()
```
mcmcgrath13, line 127: same question on down vs remove columns here

containers/ecr-viewer/src/app/data/db/utils/migrate.ts
```diff
@@ -0,0 +1,114 @@
...
+
+import { MultiDirectoryMigrationProvider } from "./multiDirectoryMigrationProvider";
+
+// Empty interface used only in migrations
+interface Database {}
+
+async function runMigration(
+  db: Kysely<Database>,
+  migrationsDir: string[],
+  command: string,
```
mcmcgrath13, line 17:
```suggestion
  command: "up" | "down",
```
these are the only options, right?

containers/ecr-viewer/integration/data/db/schema/core.test.ts
```diff
@@ -0,0 +1,449 @@
...
+  timestamp?: string;
+}
+
+describe("Common Schema: ", () => {
+  afterAll(async () => {
+    await db.destroy();
+  });
+
+  it("has a schema", async () => {
+    expect(await utils.schemaExistsByName(db, schema)).toBe(true);
```
mcmcgrath13, line 21: how are these actually using hte migration? I feel like I'm missing a step. Also can the helpers to create tables and such be replaced with the migrators now?

containers/ecr-viewer/integration/data/db/schema/core.test.ts
```diff
@@ -0,0 +1,449 @@
+import { dbNamespace, getDb } from "@/app/api/services/database";
+import { getSql } from "@/app/api/services/dialects/common";
+import { getDbUtils } from "@/app/data/db/utils";
+import { sql } from "kysely";
+
+const db = getDb();
+const schema = dbNamespace();
+let utils = getDbUtils();
```
mcmcgrath13, line 8: why not const?

containers/ecr-viewer/src/app/data/db/schemas/extended/19700101000001_initial.ts
akasper: "requires" is the wrong word for sure. I'm not sure if we can use `Common` in this context; I'll experiment. 

Lazy @akasper wants to use `<any>` because I'm special and shouldn't have to worry about this given the multiple schemas and blah, blah, blah. At one point, we use an empty `Database` interface type instead of `any`, and that makes warnings go away. 

I'm willing to rethink how to do this. Given that migrations operate in their own universe

containers/ecr-viewer/docker-compose.yaml
```diff
@@ -51,9 +51,8 @@ services:
     ports:
       - "5432:5432"
     volumes:
-      - ./seed-scripts/sql/01-init.sql:/docker-entrypoint-initdb.d/01-init.sql
```
akasper, line 54: @shanice-skylight and @rin-skylight I invite your comments on the changes to this file.

containers/ecr-viewer/api-documentation.md
```diff
@@ -96,3 +96,54 @@ curl --location '<DIBBS_URL>/ecr-viewer/api/process-zip' \
...
+
+**Permissions required** : None
+
+### Example
+
+Migrate a DB to the latest state.
+
+```sh
+curl --location '<DIBBS_URL>/ecr-viewer/api/migrate-db' \
+--form 'migration_secret=2889aa35-39d4-4bf3-a69b-aec69be1dbba'
```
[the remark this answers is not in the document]
mcmcgrath13: But given this is user facing documentation, I was thinking we want them to use something "real", which is why I put in a random UUID as that's what this should look like for them. Or at least, I assume this is user facing? Is it?

PR #280, opened 2025-02-10
feat(ecr-viewer): Add Keysely DB adapter

akasper:
# PULL REQUEST

## Summary

Add Kysely ORM to ECR Viewer, unifying DB interactions

Highlights:
* Use Kysely dialects to establish our database connections
  * unify on the `DATABASE_URL` environment variable for all dialects
* Start to move things into `Common/Core/Extended` paradigm to dry things out where we can
* Move over queries to using kysely as much as possible
  * added `getSql` to smooth over areas where the dialects don't play nice
* Update metadata bundles to reflect the fact we only ever get strings from the message parser
* Add integration tests (including heavy poaching of heavily mocked unit tests and backing them with a real db)
  * Run tests on both DBs in CI
* Run seed data conversion on both DBs for more safety nets

## Related Issue

Fixes #30 

## Acceptance Criteria

The system behaves as previously, both in MSSQL-based and PostgreSQL-based environments

## Checklist

- [ ] Specs pass
- [ ] Spot-check coverage in an environment configured to use MS SQL server as the back end
- [ ] Spot-check coverage in an environment configured to use PostgreSQL as the back end

PR #547, opened 2025-03-20
feat: Support DATABASE_URL env var for all supported database dialects

akasper:
# PULL REQUEST

_(feat) Support `DATABASE_URL` for all dialects_

## Summary

_In order to adhere to IT best practices
As an eCR Viewer system administrator
I want to limit the number of environment variables I need to allow and configure_

Adds support for using `DATABASE_URL` (and only ever using `DATABASE_URL`) for connecting to the underlying database. 

## Additional Information

Note that this targets the branch for [#280](https://github.com/CDCgov/dibbs-ecr-viewer/pull/280), and not `main`. 

## Checklist

- [ ] Try it. You'll like it. Try it. You'll like it.


containers/ecr-viewer/src/app/api/services/dialects/postgres.ts
```diff
@@ -1,15 +1,17 @@
 import { PostgresDialect } from "kysely";
 import { Pool } from "pg";
+import pdu, { DatabaseConfig } from "ts-parse-database-url";
```
BobanL, line 3: Could we just use the connectionString attribute in pool rather than adding these deps? 

https://node-postgres.com/features/connecting#connection-uri

containers/ecr-viewer/src/app/api/services/dialects/sqlserver.ts
```diff
@@ -1,6 +1,9 @@
 import { MssqlDialect } from "kysely";
 import * as tarn from "tarn";
 import * as tedious from "tedious";
+import parse from "mssql-connection-string";
+
+const dbConfig = parse(process.env.DATABASE_URL || "");
```
mcmcgrath13, line 6:
```suggestion
const { password, user, database, port, server, ...others } = parse(process.env.DATABASE_URL || "");
```

is this actually valid js? not sure, but I wonder if there's another place to shove extra config? @BobanL do you know about this oen?

BobanL: This library in general looks like to return a very small subset of possible attributes in that url. 

https://github.com/nsc-c-park/mssql-connection-string/blob/master/src/mssql-connection-string.js#L23C9 
vs
https://github.com/tediousjs/node-mssql/blob/master/lib/base/connection-pool.js#L124

containers/ecr-viewer/src/app/api/services/dialects/postgres.ts
akasper: Sure!

containers/ecr-viewer/src/app/api/services/dialects/sqlserver.ts
akasper: There are nearly 100 values we would need to support if we wanted to make a database URL work for MSSQL. Specifically, this is not an expectation that is supported by the SQL Server community. Contrast that with postgres and MySQL, which have, since their outsets, explicitly had the goal of supporting URL-based connection strings. 

I suggest that for now, in order to get this unblocked, we only support the attributes that are already supported in main. There _is not_ and off-the-shelf TypeScript library that supports connecting to MS SQL with an arbitrary URL. 

@mcmcgrath13 and @BobanL, If you have a specific fix you'd like to suggest, I'm open to it. As a matter of completing this task, I suggest we move forward with what's already there. Thoughts?

mcmcgrath13: Looking at the library internals, I don't see any way to pass through the connection string as is. @BobanL do you know which attributes users needed?

mcmcgrath13: switching to the tedious mssql connection parser seems like a fine idea either way, but doesn't really solve us as being in the middle of mapping the parsed outputs to reasonable places

mcmcgrath13: @BobanL @akasper I went ahead and switched over to the `mssql` connection parser and shoved everything I could find in a place that seemed like where it could go. I'm sure we'll find gaps in the future, but I vote we deal with it when it becomes a problem

BobanL: That sounds like the best idea!

PR #561, opened 2025-03-24
fix: handle more phone number formats and fail more gracefully

### LINKED ISSUE #497: facility contact with extension coming over as invalid
## Description

When there is an extension on the telephone number, it's displaying as invalid

## Steps to reproduce

1. Load an eCR with a phone number with an extension, e.g. `+1-608-271-9000-x11`

## Expected Result

This data should be converted. Should also handle the case without a hyphen before the extension (i.e., `608-271-9000 x11`, `608-271-9000x11`, `608-271-9000 ext 11`)

## Actual result

It is registering as invalid
Image

## Priority

#### Priority

## Notes

Design question - if it's invalid, should we show something like "Invalid number (+1-608-271-9000-x11)"

mcmcgrath13:
# PULL REQUEST

## Summary

Handle phone number with extensions (or extension like formatting). When we can't format a number instead of only showing `Invalid Number`, show `Invalid Number: <unformatted phone number>`

## Related Issue

Fixes #497

### DIFF
--- a/containers/ecr-viewer/src/app/services/formatService.ts
+++ b/containers/ecr-viewer/src/app/services/formatService.ts
@@ -165,9 +165,9 @@ export const formatCurrentAddress = (
   return formatAddress(address);
 };
 
-const VALID_PHONE_NUMBER_REGEX = /^\d{3}-\d{3}-\d{4}$/;
+const VALID_PHONE_NUMBER_REGEX = /^\d{3}-\d{3}-\d{4}( \D\w*)?$/;
 /**
- * Formats a phone number into a standard format of XXX-XXX-XXXX.
+ * Formats a phone number into a standard format of XXX-XXX-XXXX x123.
  * @param phoneNumber - The phone number to format.
  * @returns The formatted phone number or "Invalid Number" if the input is invalid or undefined if the input is empty.
  */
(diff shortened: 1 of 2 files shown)


containers/ecr-viewer/src/app/services/formatService.ts
```diff
@@ -165,9 +165,9 @@ export const formatCurrentAddress = (
   return formatAddress(address);
 };
 
-const VALID_PHONE_NUMBER_REGEX = /^\d{3}-\d{3}-\d{4}$/;
+const VALID_PHONE_NUMBER_REGEX = /^\d{3}-\d{3}-\d{4}( \D\w*)?$/;
 /**
- * Formats a phone number into a standard format of XXX-XXX-XXXX.
+ * Formats a phone number into a standard format of XXX-XXX-XXXX x123.
  * @param phoneNumber - The phone number to format.
  * @returns The formatted phone number or "Invalid Number" if the input is invalid or undefined if the input is empty.
```
mcmcgrath13, line 172:
```suggestion
 * @returns The formatted phone number or "Invalid Number: <unformmated phone number>" if the input is invalid or undefined if the input is empty.
```

angelathe (approved):

PR #565, opened 2025-03-24
chore: remove user assistance template

emmastephenson:
# PULL REQUEST

Remove user assistance template - we don't want this information to leak accidentally. Will use the Zenhub issues going forward.

## Summary

Remove user assistance template


mcmcgrath13 (approved):
🪓

PR #569, opened 2025-03-25
ci: pass ref_name during create release job

BobanL:
# PULL REQUEST

## Summary

- Improve release process by passing ref_name rather than just using the default (main)
- Use `gh run watch` to report status and add logs to workflow
  - Example run: https://github.com/BobanL/dibbs-ecr-viewer/actions/runs/14067677447/job/39394141741
 
## Related Issue

- run-container-workflow would create jobs that were cancallable by merges on main since it was using the `main` ref


mcmcgrath13 (approved):
🤖

PR #628, opened 2025-04-07
Revert "[pre-commit.ci] pre-commit autoupdate"

mcmcgrath13:
Reverts CDCgov/dibbs-ecr-viewer#625

This autoupdate took us out of sync with our actual package. They should be handled at the same time. 4.0.0-alpha isn't also a proper release, so I think we should wait for a stable release


angelathe (approved):
👍

PR #629, opened 2025-04-08
fix: make extended actually extend core

### LINKED ISSUE #417: Make `EXTENDED` schema extend `CORE`
## What needs to be done

- Ensure `EXTENDED` schema extends the `CORE` schema rather than copying the fields

## Why it needs to be done

- CORE currently has fields that are not part of extended, this leads to harder to manage code and potentially unused fields

## Acceptance Criteria

- [ ] Remove unused fields from `CORE`
- [ ] Extend the `CORE` schema to build `EXTENDED`

mcmcgrath13:
# PULL REQUEST

## Summary

Make the `extended` schema actually extend the `core` schema. This allows us to get rid of `Common` and just use `Core` inside the viewer (saving still needs to know about both, but listing doesn't).

Because the extended schema is in use and the core schema is not, the fields where they differed were consolidated to the extended version (e.g. core had `patient_name_first`, extended has `first_name`, now core has `first_name` and extended uses that).

## Related Issue

Fixes #417

## Acceptance Criteria

- [ ] Remove unused fields from `CORE`
- [ ] Extend the `CORE` schema to build `EXTENDED`

## Additional Information

While I'd say it's generally a bad idea to edit migrations instead of adding new ones, because this a fast follow and users end up in the same place as they started, it seems simpler to gloss over this so we're setting ourselves up for simpler development long run


containers/ecr-viewer/integration/helpers/ddl.ts
```diff
@@ -13,11 +13,19 @@ export const dropExisting = async () => {
...
+/**
+ * Builds the core schema to a test database
+ */
+export const buildCore = async () => {
+  process.env.METADATA_DATABASE_SCHEMA = "core";
+  await migrateUp();
+};
+
 /**
  * Clears the common schema tables
```
mcmcgrath13, line 25:
```suggestion
 * Clears the core schema tables
```

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/19700101000000_initial.ts
```diff
@@ -1,56 +1,83 @@
...
+      .execute();
+  }
 }
 
 /**
- * Based on ecr-viewer/sql/core.sql.
- * Core schema tear down. This version is hard-coded to the
- * original postgres implementation. Future versions will be
- * database-agnostic.
+ * Common schema initialization.
```
austin-hall-skylight, line 74:
```suggestion
  * Based on ecr-viewer/sql/core.sql.
 * Core schema tear down. This version is hard-coded to the
 * original postgres implementation. Future versions will be
 * database-agnostic.
```

austin-hall-skylight (approved):

PR #631, opened 2025-04-08
fix: Make patient age field dynamic

### LINKED ISSUE #508: Change "Current Age" to "Age at Encounter"
## What needs to be done
A patient's age is calculated at the time of their encounter. Change the header in Patient Info from "Current Age" To "Age at Encounter"

### LINKED ISSUE #626: Remove save-fhir-data endpoint
## What needs to be done

`save-fhir-data` is no longer needed for the ecr-viewer so can be removed. There are `orchestration` configs that use the endpoint and they will also need to be edited.

## Why it needs to be done

This is the next step in disentangling the ecr-viewer from orchestration. In addition it allows us to remove an endpoint that will then not need to be secured.

## Acceptance Criteria

- [ ] the `/save-fhir-data` endpoint in the ecr-viewer no longer exists.
- [ ] the relevant orchestration configs are updated to not use the `save-fhir-data` endpoint
- [ ] The relevant orchestration configs return the bundle.
- [ ] All of the orchestration code for using the `/save-fhir-data` endpoint is removed.
- [ ] Any other integration tests that use `/save-fhir-data` are either removed or changed.

JNygaard-Skylight:
# PULL REQUEST

## Summary

Makes the patient age field dynamic:
- if there is an date of death:
  - name: Age at Death
  - value: age at death
  - tooltip: none
- If there is an Encounter start date
  - name: Age at Encounter
  - value: age at start of encounter
  - tooltip: none
image
- If there is not an encounter start date and encounter end date in the past
  - name: Age at Encounter
  - value: age at the end of the encounter
  - tooltip: "Age at end date of encounter. Start date of encounter is not available."
image
- If there is not an encounter start date, and the encounter end date is in the future
  - name: Age at Encounter
  - value: age at the eCR created date.
  - tooltip: "Using the date eCR was created as a proxy for date of encounter. No encounter start date and encounter end date is in the future."
image
- If there is not an encounter period
  - name: Age at Encounter
  - value: age at the eCR created date.
  - tooltip: "Using the date eCR was created as a proxy for date of encounter. No encounter date available."
image
- If there is no date of birth
  - Do not show the patient's age.

In addition to the above, always show the onset age, even if there is a death date.

## Related Issue

Fixes #508

### DIFF
--- a/containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
+++ b/containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
@@ -699,7 +699,7 @@ export const createPatientAgeDataProp = (
   );
   const patientDOBString = evaluateOne(fhirBundle, fhirPathMappings.patientDOB);
 
-  let title = "Current Age";
+  let title = "Age at Encounter";
   let toolTip;
   let value;
 
@@ -723,7 +723,6 @@ export const createPatientAgeDataProp = (
 
   // Handle encounter start date
   if (encounterPeriod?.start) {
-    title = "Age at Encounter";
     value = formatAge(calculatePatientAge(fhirBundle, encounterPeriod.start));
     return { title, toolTip, value };
   }
@@ -733,15 +732,19 @@ export const createPatientAgeDataProp = (
     const encounterEnd = DateTime.fromJSDate(new Date(encounterPeriod.end));
 
     if (encounterEnd <= DateTime.now()) {
-      title = "Age at Encounter";
       toolTip =
         "Age at end date of encounter. Start date of encounter is not available.";
       value = formatAge(calculatePatientAge(fhirBundle, encounterPeriod.end));
     } else {
-      value = formatAge(calculatePatientAge(fhirBundle));
+      value = formatAge(
+        calculatePatientAge(
+          fhirBundle,
+          evaluateOne(fhirBundle, fhirPathMappings.dateTimeEcrCreated),
+        ),
+      );
       if (value) {
         toolTip =
-          "Using the date eCR was received as a proxy for date of encounter. No encounter start date and encounter end date is in the future.";
+          "Using the date eCR was created as a proxy for date of encounter. No encounter start date and encounter end date is in the future.";
       }
     }
     return { title, toolTip, value };
--- a/containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
+++ b/containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
@@ -355,7 +355,7 @@ export const evaluateDemographicsData = (fhirBundle: Bundle) => {
       value: evaluatePatientDOB(fhirBundle),
     },
     {
-      title: "Current Age",
+      title: "Age at Encounter",
       value: formatAge(calculatePatientAge(fhirBundle)),
     },
     {
--- a/containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
+++ b/containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
@@ -147,51 +147,42 @@ export interface Age {
 }
 
 /**
- * Calculates the age of a patient to a given date or today, unless DOD exists.
+ * Calculates the patient's age at a specific point in time or the current date.
  * @param fhirBundle - The FHIR bundle containing patient information.
- * @param [givenDate] - Optional. The target date to calculate the age. Defaults to the current date if not provided.
- * @returns - The exact age of the patient in years/months/days, or undefined if date of birth is not available or if date of death exists.
+ * @param [givenDate] - Optional date to calculate age at, overrides encounter dates
+ * @returns The patient's age in years, or undefined if:
+ *   - Patient has a recorded death date
+ *   - Patient has no birth date
  */
 export const calculatePatientAge = (
   fhirBundle: Bundle,
   givenDate?: string,
 ): Age | undefined => {
-  const deathDate = evaluateOne(fhirBundle, fhirPathMappings.patientDOD);
+  const patientDOBString = evaluateOne(fhirBundle, fhirPathMappings.patientDOB);
 
-  // if a death date is available, don't calculate patient age
+  // If no patient DOB is available, return undefined.
+  if (!patientDOBString) {
+    return undefined;
+  }
+
+  const patientDOB = DateTime.fromJSDate(new Date(patientDOBString));
+
+  // If date is provided by caller, use that.
+  if (givenDate) {
+    return calculateYearsFrom(
+      DateTime.fromJSDate(new Date(givenDate)),
+      patientDOB,
+    );
+  }
+
+  // If a death date is available, return undefined.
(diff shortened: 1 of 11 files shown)


mcmcgrath13 (approved):
👴

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
```diff
@@ -355,7 +355,7 @@ export const evaluateDemographicsData = (fhirBundle: Bundle) => {
       value: evaluatePatientDOB(fhirBundle),
     },
     {
-      title: "Current Age",
+      title: "Age at Encounter",
       value: formatAge(calculatePatientAge(fhirBundle)),
```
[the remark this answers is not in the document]
angelathe: @JNygaard-Skylight Sorry! Removed this from the merge queue until we sort this out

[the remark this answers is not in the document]
JNygaard-Skylight: The logic for this is more complicated than I thought. The function `calculatePatientAge` checks the following conditions, returning the first one it satisfies:
1) If there is a death date return `undefined`.
2) If there is no date of birth return `undefined`.
3) If a date is passed in as a parameter get the age at that date. Currently we only do this with age at condition onset
4) If there is an encounter date, return age at encounter
5) Return the current age.

So to summarize: The patient age shown in the demographics section is the age at encounter, unless, in the rare situation where there is not an encounter date, it will be the current age. 

A question for @ashton-skylight: Is the above summarized behavior what we want, and if it is should we denote this somehow? We could change the label depending what age we are showing?

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
```diff
@@ -222,7 +213,10 @@ export const calculatePatientAgeAtDeath = (
  * @param earlierDate DateTime earlier in time
  * @returns An `Age`
  */
-const getPatientAge = (laterDate: DateTime, earlierDate: DateTime): Age => {
+const calculateYearsFrom = (
```
mcmcgrath13, line 216: this doesn't strictly calculate years - is there a better name here?

containers/ecr-viewer/src/app/services/formatDateService.ts
```diff
@@ -1,3 +1,5 @@
+import { Period } from "fhir/r4";
```
mcmcgrath13, line 1: can you check this doesn't mess with bundle size - I believe this file should be client side-able and that's why it's split out from the other fhir formatter functions

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
mcmcgrath13: maybe `Duration`?

containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
```diff
@@ -316,13 +316,9 @@ const _fhirPathMappings: { [K in FhirPathKeys]: Omit<FhirPath<K>, "name"> } = {
...
 
   // Encounter Info
-  encounterEndDate: {
-    type: "string",
-    path: "Bundle.entry.resource.where(resourceType = 'Encounter').period.end",
-  },
-  encounterStartDate: {
-    type: "string",
-    path: "Bundle.entry.resource.where(resourceType = 'Encounter').period.start",
+  encounterPeriod: {
```
mcmcgrath13, line 319: ❤️

akasper (approved):
- [x] Searched the entire codebase for instances of `encounterStartDate` and `encounterEndDate` to validate that they had all been replaced. (They have.)
- [x] Validated tests and ran local tests to verify correctness of behavior for `calculateYearsFrom`. (Everything worked, but I've provided suggestions for additional test coverage.)
- [x] Validated updated signatures for `formatStartEndDateTime` and `formatStartEndDate`. Verified that calls using the old format were modified throughout the codebase. 
- [x] Investigated changes to tests for `calculatePatientAge`, `createPatientAgeDataProp`, and `calculatePatientAgeAtDeath`. (They look good.)
- [x] Searched the codebase for usage of the old format of FHIR path mappings. (Didn't find any.)
- [x] Visually verified correctness of UI, albeit using a small local dataset. 

@mcmcgrath13 has a few suggestions -- in particular the suggested rename of `calculateYearsFrom` -- but beyond those and my test suggestions (not requirements), this looks good to me.

containers/ecr-viewer/src/app/services/formatDateService.ts
JNygaard-Skylight: Thinking about it some more, I think I like the idea that the formatting services should not/do not need to know about FHIR objects. I'll rework my changes so that the `formatStartEnd` can handle `Period`s, without knowing about them explicitly, using the power of interfaces.

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
```diff
@@ -735,3 +724,64 @@ export const evaluatePatientLanguage = (fhirBundle: Bundle) => {
...
+  // If patient has death date, return empty object
+  if (hasDeathDate) {
+    return { title, toolTip, value };
+  }
+
+  // Handle encounter start date
+  if (encounterPeriod?.start) {
+    title = "Age at Encounter";
+    value = formatAge(calculatePatientAge(fhirBundle, encounterPeriod.start));
+    return { title, toolTip, value };
```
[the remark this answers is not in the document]
JNygaard-Skylight: My thinking is that this is the default behaviour, and it does exactly what one would expect, so there is no additional information to give.

containers/ecr-viewer/src/app/tests/services/evaluateFhirDataServices.test.ts
```diff
@@ -378,4 +386,197 @@ Home: 123-456-6909`,
       expect(actual).toEqual("");
     });
   });
+
+  describe("Calculate Patient Age", () => {
```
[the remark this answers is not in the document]
JNygaard-Skylight: - Leap years are the type of thing I would expect the date time library we are using to account for. If you think it is important I can add a unit test for it, but I do not think it is needed.
- While I think it is technically possible, I think we can safely assume there will only be `Patient`. We have not defined logic for multiple `Patient`s anywhere else.
- Off-by-one date issues are something I would leave to the datetime library.
- The only change for infants is how we format the age, so that should be covered by the `formatAge` unit tests.
- Added more robust logic for determining if a patient is deceased.

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
JNygaard-Skylight: I went with `calcuateAge` as it returns an `Age` object, does that feel better?

JNygaard-Skylight: I went with `calcuateAge` as it returns an `Age` object, does that feel better?

PR #632, opened 2025-04-08
fix: handle rollbacks on partially failed save

### LINKED ISSUE #69: Add rollback functionality when saving FHIR bundle with metadata
## What needs to be done

- Add rollback functionality to handle if either FHIR bundle or metadata save fail.

## Why it needs to be done

- Currently either the FHIR bundle save or metadata save steps could fail without rolling back the other

## Acceptance Criteria

- When `saveFhirData` then no metadata should be saved
- When `saveFhirMetadata` then no bundle should be saved

## Technical Notes

- Consider using [`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) instead of `Promise.all` to check if either process failed.

## Additional context

[OPTIONAL, delete if not needed] 

mcmcgrath13:
# PULL REQUEST

## Summary

Handle rolling back fhir- or meta- data if the other fails when trying to save both.

Highlights:
* Kysely transactions roll back when an error is thrown, so check if fhir saving failed inside the transaction and throw an error if it does
* If the database saving throws an error, check if it was before rolling back the fhir data and if it was, roll back the fhir data
* Check if an eCR already exists and bail out early with a nicer error message than a 500
* move the blob saving/deleting code to the client files

## Related Issue

Fixes #69

## Acceptance Criteria

- [ ] When saveFhirData then no metadata should be saved
- [ ] When saveFhirMetadata then no bundle should be saved

### DIFF
--- a/containers/ecr-viewer/jest.config.js
+++ b/containers/ecr-viewer/jest.config.js
@@ -20,6 +20,7 @@ const customJestConfig = {
     process.env.TEST_TYPE === "integration"
       ? ["<rootDir>/integration/setup.ts"]
       : [],
+  // transformIgnorePatterns: ["/node_modules/(?!(@azure)/)", "\\.pnp\\.[^\\\/]+$"]
 };
 
 // createJestConfig is exported this way to ensure that next/jest can load the Next.js config which is async
--- a/containers/ecr-viewer/jest.config.js
+++ b/containers/ecr-viewer/jest.config.js
@@ -20,7 +20,6 @@ const customJestConfig = {
     process.env.TEST_TYPE === "integration"
       ? ["<rootDir>/integration/setup.ts"]
       : [],
-  // transformIgnorePatterns: ["/node_modules/(?!(@azure)/)", "\\.pnp\\.[^\\\/]+$"]
 };
 
 // createJestConfig is exported this way to ensure that next/jest can load the Next.js config which is async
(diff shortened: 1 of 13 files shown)


containers/ecr-viewer/jest.config.js
```diff
@@ -20,6 +20,7 @@ const customJestConfig = {
     process.env.TEST_TYPE === "integration"
       ? ["<rootDir>/integration/setup.ts"]
       : [],
+  // transformIgnorePatterns: ["/node_modules/(?!(@azure)/)", "\\.pnp\\.[^\\\/]+$"]
```
austin-hall-skylight, line 23: Should this be uncommented or deleted?

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
```diff
@@ -21,160 +23,53 @@ interface SaveResponse {
...
+    return await saveToAzure(body, objectKey);
+  } else if (saveSource === GCP_SOURCE) {
+    return await saveToGCP(body, objectKey);
+  } else {
     return {
-      message: "Failed to save FHIR bundle.",
-      status: 500,
+      message:
+        'Invalid save source. Please provide a valid value for \'saveSource\' ("s3", "azure", or "gcp").',
+      status: 400,
```
austin-hall-skylight, line 50: I like this change.

austin-hall-skylight (approved):

PR #634, opened 2025-04-10
fix: consistently format units in reference ranges

### LINKED ISSUE #103: BUG: Fix units display on lab ref ranges
## Description

Please update how units are displayed for lab reference ranges. If the unit is `%`, we want no visible space between the value and the % -- `<value>%`. For all other cases, units should be separated from the value by a space like `<value> <unit>`.

To do this, we have to change the FHIR path for `observationRefRange`.  Currently, we pull all the reference range values from the `text` field. If the `low` and `high` value and unit fields exist under `referenceRange`, we should pull from them instead. 

## Steps to reproduce

`10e5b07b-5d3f-426f-a7e1-13ea909ff5a5`: [Link to eCR Viewer](http://localhost:3000/view-data?id=32#lab-results-from-mercy-campus-muskegon-mi-mcmu-hospital-lab)

## Priority

#### Priority

## Notes

No FHIR conversion should be required for this ticket.

mcmcgrath13:
# PULL REQUEST

## Summary

Format observation reference ranges consistent with other quantities (when low/high available)

## Related Issue

Fixes #103

### DIFF
--- a/containers/ecr-viewer/src/app/services/formatService.ts
+++ b/containers/ecr-viewer/src/app/services/formatService.ts
@@ -6,6 +6,7 @@ import {
   ContactPoint,
   HumanName,
   PatientContact,
+  Quantity,
   RelatedPerson,
 } from "fhir/r4";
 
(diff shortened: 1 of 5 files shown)


containers/ecr-viewer/src/app/services/formatService.ts
```diff
@@ -325,6 +327,51 @@ export const formatCodeableConcept = (
...
+ * @returns formatted string
+ */
+export const formatRange = (data: Range | undefined): string | undefined => {
+  if (!data) return;
+  const low = formatQuantity(data.low);
+  const high = formatQuantity(data.high);
+  if (low && high) {
+    return `${low} - ${high}`;
+  } else if (low) {
+    return `>=${low}`;
```
angelathe, line 369: Could `<low>` and `<high>` also apply to dates, and if so, does the `>=` and `<=` make sense in that case?

Though, I saw that this is currently only applied to Observation ref ranges. I would maybe make a note about this so that it won't accidentally get applied to dates in the future?

containers/ecr-viewer/src/app/services/formatService.ts
```diff
@@ -325,6 +327,51 @@ export const formatCodeableConcept = (
   return undefined;
 };
 
+// Map from computer to human readable units
+const UNIT_MAP = new Map([
```
angelathe, line 331: Just curious why you moved this from `utils` to `formatService`?

containers/ecr-viewer/src/app/services/formatService.ts
mcmcgrath13: `Range` is defined using `Quantities`, so we shouldn't run into this with dates as that would use a `Period` (pretty sure)

angelathe (approved):
💯

PR #635, opened 2025-04-10
fix: don't record exceptions for handled orchestration errors

### LINKED ISSUE #456: Orchestration returning internal server error if config does not exist
## Description

- `/proccess-zip` and `/process-message` both return a 500 internal server error if invalid configs are provided

## Steps to reproduce

1. `POST` against `/process-zip` or `/process-message` with config of `garbage.json`

## Expected Result

- A 400 should be returned

## Actual result

- 500 is returned

## Priority

#### Priority

## Notes

- 
image
Throws on L314
Image

mcmcgrath13:
# PULL REQUEST

## Summary

Per the [open telemetry docs](https://opentelemetry.io/docs/specs/otel/trace/exceptions/) `An exception SHOULD be recorded as an Event on the span during which it occurred if and only if it remains unhandled when the span ends and causes the span status to be set to ERROR`. However, we were recording an exception in a case we were handling, which was causing the case to 500 instead of return the response that was constructed.

## Related Issue

Fixes #456


akasper (approved):
The code changes make perfect sense. 

The only risk I can imagine is that some downstream telemetry is somehow relying upon the old way of handling the `FileNotFound`. Even if that is the case, though: That 1) goes against the OpenTelemetry spec and 2) is something we'll hear about right away. 

I didn't see anything in any of our docs explaining why this was the way it was, so I'm confident that this is fine and (in the worst case) we'll know pretty quickly if it's not.

PR #647, opened 2025-04-18
feat: add optional flag to return bundle from process-zip

### LINKED ISSUE #648: Add optional flag to get bundle from `process-zip`

mcmcgrath13:
# PULL REQUEST

## Summary

Add an optional flag to the `process-zip` endpoint to enable returning the processed FHIR bundle

## Related Issue

Fixes #648

### DIFF
--- a/containers/ecr-viewer/api-documentation.md
+++ b/containers/ecr-viewer/api-documentation.md
@@ -1,4 +1,6 @@
-# View eCR
+# eCR Viewer API Documentatino
+
+## View eCR
 
 Display an eCR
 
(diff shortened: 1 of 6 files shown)


angelathe (approved):
Looks good! (pending if there are any documentation updates)

mcmcgrath13:
> Are there any docs that need to get updated with this new flag?

I don't think there are any docs on this 😬 (at least in-repo), so no..., unless @emmastephenson did you build out some docs on this recently?

emmastephenson:
> > Are there any docs that need to get updated with this new flag?
> 
> I don't think there are any docs on this 😬 (at least in-repo), so no..., unless @emmastephenson did you build out some docs on this recently?

@mcmcgrath13 I do indeed have many out-of-repo docs 😅 (Which we might want to move in-repo??) Hate to ask, but it might be good to have a source-of-truth doc about this endpoint in the repo - thinking we could update [this doc](https://github.com/CDCgov/dibbs-ecr-viewer/blob/main/containers/ecr-viewer/api-documentation.md) and just give a list of params, an example request, and maybe an expected response?

If that feels out of scope for this ticket, totally fine, we can add it later!

PR #656, opened 2025-04-24
feat: make `extended` the default schema, any schema on any db 

### LINKED ISSUE #651: Make `extended` schema default, add/leverage metadata schema env var to opt in to `core`
Currently, we use the config's metadata db as a proxy for extended/core (sqlserver=extended and pg=core). We want to change this to make the extended schema the default if a db type is provided and let core be opted into via the already existing (though not set by user) `METADATA_DATABASE_SCHEMA` env var.

This will likely be a small code change, but may require changes to tests to make sure their assumptions still hold.

Note, no users are using a `PG` config to date, so this can be merged as a minor change and not a major breaking change

mcmcgrath13:
# PULL REQUEST

## Summary

Make `extended` the default schema. Allow users to manually set `METADATA_DATABASE_SCHEMA` to `core` if they want the smaller subset. This also means we officially support `core` and `extended` on both `sqlserver` and `pg`!

Note: By the letter of the law, this is a breaking change, but by the spirit, this is not as no one is currently using the core schema

## Related Issue

Fixes #651


akasper:
This all makes perfect sense.

gordonfarrell (approved):
LGTM!

containers/orchestration/docker-compose.yaml
```diff
@@ -26,6 +26,7 @@ services:
     profiles: ["", "azure", "aws", "sqlserver"]
     environment:
       - CONFIG_NAME=AZURE_PG_NON_INTEGRATED
+      - METADATA_DATABASE_SCHEMA=core
```
angelathe, line 29: Can you explain why we want to set this specifically to `core`?

angelathe (approved):
TY!

containers/orchestration/docker-compose.yaml
mcmcgrath13: this compose I believe is only used for testing (based on the hard coded config name above) and the tests relied on the old postgres==core paradigm, so I added in the environment var so the tests would still work as intended

PR #658, opened 2025-04-24
ci: disable automatic release drafter workflow trigger

gordonfarrell:
# PULL REQUEST

## Summary

Release drafter is not functioning as expected and is opening dozens and dozens of draft releases, disabling while we look into it


JNygaard-Skylight (approved):

.github/workflows/release-drafter.yaml
```diff
@@ -1,12 +1,13 @@
...
-  # pull_request event is required only for autolabeler
-  pull_request:
-    types: [opened, reopened, edited, synchronize]
+  workflow_dispatch:
+#  push:
+#    branches:
+#      - main
+# pull_request event is required only for autolabeler
+#  pull_request:
+#    types: [opened, reopened, edited, synchronize]
```
mcmcgrath13, line 10: I think we still want the pull request labels to be added

mcmcgrath13:
looking at the repo, pinning our version to `6.0.0` will likely fix the issue. I'm still not fully clear on whether we want pre-releases or not and what the implications are there

Issue #513, opened 2025-03-14
Add fk on `ecr_rr_conditions` table to the `condition_reference` table


mcmcgrath13:
## What needs to be done

With the addition of the `condition_reference` table in #512 , we'll have a consistent identifier for conditions. We need to add a column to the `ecr_rr_conditions` table as a fk to that table

## Why it needs to be done

We'll be using the conditions for authorization and filtering, so we need a pointer from the ecr condition to the source condition so we can implement this.

## Acceptance Criteria

_List out all expected functionality/requirements_

- A nullable column called `condition_id` should be added to the `ecr_rr_conditions` table (both schemas). It should be a foreign key to `condition_reference`
  - postgres
  - sql server
- On saving data to `ecr_rr_conditions` it should be stamped with this ID if it can be determined

## Technical Notes

_Owned by eng team - write out any important technical considerations such as: architecture plan/notes, data dependencies, testing plan, etc. Also link any tickets that are dependent on or related to this task_

- Because we're fundamentally dealing with strings, the stamping will be best effort and can be refined as time goes on

## Additional context

[OPTIONAL, delete if not needed]
_Include any additional context or information here that doesn't fit into other parts of the template_

akasper:
https://github.com/CDCgov/dibbs-ecr-viewer/tree/akasper/condition-ref-fk

mcmcgrath13:
> https://github.com/CDCgov/dibbs-ecr-viewer/tree/akasper/condition-ref-fk

could you open a draft PR?

akasper:
@mcmcgrath13: https://github.com/CDCgov/dibbs-ecr-viewer/pull/678

PR #678, opened 2025-05-01
feat: Add foreign key on `ecr_rr_conditions` to the `condition_reference` table

akasper:
# PULL REQUEST

Now that we have a consistent identifier for conditions (via the `condition_reference` table), are add a foreign key reference to those conditions in our per-eCR reportability response conditions. 

## Summary

* A database schema change that adds `condition_id` to `ecr_rr_conditions`, as well as a foreign key
* The attendant TypeScript interface definition changes for `ecr_rr_conditions`
* Modifications to data saving logic
* Logic for backfilling existing data
* Tests that document and prove the functionality

## Related Issue

Ticket #513 

## Acceptance Criteria

- A nullable column called `condition_id` should be added to the `ecr_rr_conditions` table (both schemas). It should be a foreign key to `condition_reference`
  - postgres
  - sql server
- On saving data to `ecr_rr_conditions` it should be stamped with this ID if it can be determined
- Create a migration such that the data already in the ecr_rr_conditions table gets an accurate fk reference to the condition_reference table


containers/ecr-viewer/src/app/api/migrate-db/migrations/core/20250430160000_condition_fk.ts
```diff
@@ -0,0 +1,86 @@
...
+          LIMIT 1;
+        END IF;
+        RETURN NEW;
+      END;
+      $$ LANGUAGE plpgsql;
+    `.execute(_db);
+
+    await sql`
+      CREATE TRIGGER trg_set_condition_code
+      BEFORE INSERT ON ${sql.raw(schema)}.ecr_rr_conditions
```
mcmcgrath13, line 47: instead of doing this with triggers can we just do an update on the existing table and instead update the metadata parser to just pull out the code from the fhir bundle so we get the original source of truth from what was there and then save that as we do regular data for go forward?

containers/ecr-viewer/src/app/api/save-fhir-data/service.ts
```diff
@@ -308,13 +308,25 @@ const saveRR = async (
...
     const rr_conditions_uuid = randomUUID();
+    
+    const rr_code = await trx
+      .selectFrom("condition_reference")
+      .select("code")
+      .where("condition_name", "=", rrItem.condition)
+      .executeTakeFirst();
+
+    if (!rr_code) {
+      console.error(`Condition ${rrItem.condition} not found in condition_reference table`);
```
mcmcgrath13: can we add this to the metadata instead? Our conditions reference comes from TRS, so we should be able to just add the code to the metadata schema and get it from the fhir bundle if it exists

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/index.ts
```diff
@@ -2,5 +2,7 @@ export default {
   "19700101000000_initial.ts": require("./19700101000000_initial"),
   "20250425111400_user_prog.ts": require("./20250425111400_user_prog"),
   "20250428110000_condition_ref.ts": require("./20250428110000_condition_ref"),
+  "20250430160000_condition_fk.ts": require("./20250430160000_condition_fk"),
```
mcmcgrath13, line 5: because 202505 has already merged, can you change the condition fk to be later? also, what's the thinking on two migrations instead of doing this in one go?

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/20250506120000_backfill_condition_codes.ts
```diff
@@ -0,0 +1,56 @@
+import { Kysely, sql } from "kysely";
+import { AnyDb } from "@/app/data/metadataDb/database";
+import { dbDialect, dbNamespace } from "@/app/data/metadataDb/utils/db-config";
+
+export async function up(db: Kysely<AnyDb>): Promise<void> {
+  const schema = dbNamespace();
+
+  if (dbDialect() === "postgres") {
+    await db
```
mcmcgrath13, line 9: this db needs to be namespaced

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/20250506120000_backfill_condition_codes.ts
```diff
@@ -0,0 +1,56 @@
...
+      FROM ecr_rr_conditions erc
+      JOIN ConditionCodeLookup ccl ON erc.uuid = ccl.uuid;
+    `.execute(db);
+  }
+}
+
+export async function down(db: Kysely<AnyDb>): Promise<void> {
+  const schema = dbNamespace();
+
+  await db
```
mcmcgrath13, line 49: same here

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/20250506120000_backfill_condition_codes.ts
```diff
@@ -0,0 +1,56 @@
...
+    // SQL Server requires a slightly different approach for correlated subqueries in UPDATE
+    // Instead, we use a CTE for clarity and compatibility
+    await sql`
+      WITH ConditionCodeLookup AS (
+        SELECT 
+          erc.uuid,
+          cr.code AS new_condition_code
+        FROM ecr_rr_conditions erc
+        JOIN condition_reference cr ON erc.condition = cr.condition_name
+        WHERE erc.condition_code IS NULL
```
mcmcgrath13, line 36: any reason not to do this with kysely? and for postgres as well? we use kyysely CTE quite a bit for the ecr listing selects

containers/ecr-viewer/src/app/api/save-fhir-data/service.ts
```diff
@@ -308,13 +308,21 @@ const saveRR = async (
   // Loop through each condition/rule object in rr array
   for (const rrItem of metadata.rr) {
     const rr_conditions_uuid = randomUUID();
+
+    const rr_code = await trx
+      .selectFrom("condition_reference")
+      .select("code")
+      .where("condition_name", "=", rrItem.condition)
+      .executeTakeFirst();
```
mcmcgrath13, line 316: still needs to be sourced from the metadata data (and metadata parser updated to add it)

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/20250509000002_backfill_condition_codes.ts
```diff
@@ -0,0 +1,55 @@
...
+            "condition_reference.condition_name",
+            "=",
+            "ecr_rr_conditions.condition",
+          )
+          .limit(1),
+      }))
+      .where("condition_code", "is", null)
+      .execute();
+  } else if (dbDialect() === "sqlserver") {
+    const rows = await _db
```
mcmcgrath13, line 26: would this branch also work for postgres?

containers/ecr-viewer/src/app/api/save-fhir-data/service.ts
akasper: ECR Viewer side of things is properly set up but getting a little lost in the weeds of orchestration. Would you mind taking a look at what I've done with the schema changes in the message parser @mcmcgrath13

containers/message-parser/app/default_schemas/core.json
```diff
@@ -59,6 +59,14 @@
...
         }
       },
+      "condition_code": {
+        "fhir_path": "iif(Observation.valueCodeableConcept.text.exists(), Observation.valueCodeableConcept.text, Observation.valueCodeableConcept.coding.display)", 
+        "data_type": "string",
+        "nullable": true,
+        "metadata": {
+          "category": "RR Reportability Information"
+        }
+      },
```
mcmcgrath13: move this up into the `rr` section and maybe just name `code`?

containers/ecr-viewer/src/app/api/save-fhir-data/types.ts
```diff
@@ -41,6 +41,7 @@ interface ruleSummary {
 
 interface RR {
   condition: string;
+  condition_code: string | undefined;
```
mcmcgrath13, line 44: this is out of sync with the schema

containers/message-parser/app/default_schemas/core.json
```diff
@@ -59,6 +59,14 @@
           "category": "RR Reportability Information"
         }
       },
+      "code": {
+        "fhir_path": "iif(Observation.valueCodeableConcept.text.exists(), Observation.valueCodeableConcept.text, Observation.valueCodeableConcept.coding.code)",
```
mcmcgrath13, line 63: we don't want the text here, we want the code that corresponds to the text

containers/message-parser/app/default_schemas/ecr.json
```diff
@@ -354,6 +354,14 @@
           "category": "RR Reportability Information"
         }
       },
+      "code": {
+        "fhir_path": "iif(Observation.valueCodeableConcept.text.exists(), Observation.valueCodeableConcept.text, Observation.valueCodeableConcept.coding.code)",
```
mcmcgrath13, line 358: same here

mcmcgrath13: though I don't think we use this schema? so you can just leave it alone. Only `core.json` and `extended.json` are in use that I know of

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/index.ts
```diff
@@ -2,5 +2,7 @@ export default {
   "19700101000000_initial.ts": require("./19700101000000_initial"),
   "20250425111400_user_prog.ts": require("./20250425111400_user_prog"),
   "20250428110000_condition_ref.ts": require("./20250428110000_condition_ref"),
-  "20250505100700_ecr_fks.ts": require("./20250505100700_ecr_fks"),
+  "20250509000000_condition_fk.ts": require("./20250509000000_condition_fk"),
+  "20250509000001_ecr_ref.ts": require("./20250509000001_ecr_ref"),
```
mcmcgrath13, line 6: this name needs to be changed back to the original - it'll mess with kyseley's tracking to change it

containers/message-parser/app/default_schemas/core.json
```diff
@@ -60,7 +60,7 @@
         }
       },
       "code": {
-        "fhir_path": "iif(Observation.valueCodeableConcept.text.exists(), Observation.valueCodeableConcept.text, Observation.valueCodeableConcept.coding.code)",
+        "fhir_path": "iif(Observation.valueCodeableConcept.coding.code.exists(), Observation.valueCodeableConcept.coding.code)",
```
mcmcgrath13, line 63: I think we still wanted that test clause to be text as the text is when we stamp, so if we didn't stamp, it didn't/won't join

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/index.ts
```diff
@@ -2,5 +2,6 @@ export default {
   "19700101000000_initial.ts": require("./19700101000000_initial"),
   "20250425111400_user_prog.ts": require("./20250425111400_user_prog"),
   "20250428110000_condition_ref.ts": require("./20250428110000_condition_ref"),
-  "20250505100700_ecr_fks.ts": require("./20250505100700_ecr_fks"),
+  "20250509000000_condition_fk.ts": require("./20250509000000_condition_fk"),
```
mcmcgrath13, line 5: rename these files to be after the `ecr_fk` - should be in merge order

containers/message-parser/app/default_schemas/core.json
akasper: Do you mean a SQL-like where?

containers/ecr-viewer/integration/saveFhirDataService/extended.test.ts
```diff
@@ -232,4 +269,37 @@ describe("saveFhirData - extended", () => {
     expect(rolledback).toBeFalse();
     expect(res).toHaveLength(0);
   });
+  it("should reference the condition_code foreign key", async () => {
```
mcmcgrath13, line 272: could you move this test closer to the other rr-ey ones?

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/20250505100700_ecr_fks.ts
mcmcgrath13, line 1: delete?

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/20250509000000_condition_fk.ts
mcmcgrath13, line 1: could you update the date here to be closer to today?

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/20250509000000_condition_fk.ts
```diff
@@ -0,0 +1,82 @@
...
+export async function up(db: Kysely<AnyDb>): Promise<void> {
+  const schema = dbNamespace();
+  const _db = db.withSchema(schema);
+
+  await _db.schema
+    .alterTable("ecr_rr_conditions")
+    .addColumn("condition_code", "varchar(20)")
+    .execute();
+
+  await _db.schema
```
mcmcgrath13, line 19: why split the fk add into a separate call instead of as part of the column def?

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/20250509000000_condition_fk.ts
```diff
@@ -0,0 +1,82 @@
...
+    const rows = await _db
+      .selectFrom("ecr_rr_conditions as erc")
+      .innerJoin(
+        "condition_reference as cr",
+        "erc.condition",
+        "cr.condition_name",
+      )
+      .select(["erc.uuid", "cr.code as new_code"])
+      .where("erc.condition_code", "is", null)
+      .execute();
```
mcmcgrath13, line 56: is there any way we can not have a split on dialect here? does the sqlserver one work for postgres?

containers/message-parser/app/default_schemas/core.json
```diff
@@ -59,6 +59,14 @@
           "category": "RR Reportability Information"
         }
       },
+      "code": {
+        "fhir_path": "iif(Observation.valueCodeableConcept.text.exists(), Observation.valueCodeableConcept.coding.code).where(Observation.valueCodeableConcept.coding.code = %snomedCode)",
```
mcmcgrath13, line 63: this fhir_path looks questionable to me. The where should typically be filtering the currently slected thing in the path and a `code` is not an `Observation`

I'd expect something more like
```
iif(Observation.valueCodeableConcept.text.exists(), Observation.valueCodeableConcept.coding.where(system = 'http://snomed.info/sct').code)
```

containers/message-parser/app/default_schemas/extended.json
```diff
@@ -388,6 +388,14 @@
           "category": "RR Reportability Information"
         }
       },
+      "code": {
+        "fhir_path": "iif(Observation.valueCodeableConcept.text.exists(), Observation.valueCodeableConcept.coding.code).where(Observation.valueCodeableConcept.coding.code = %snomedCode)",
```
mcmcgrath13, line 392: same here

containers/message-parser/app/default_schemas/core.json
mcmcgrath13: also, looking at the trigger reference service it seems like the addition of `text` doesn't necessarily indicate we got a match from TES and therefore that doesn't indicate the condition code will match up to one of our reference codes. @emmastephenson when this happens, do we want the saving of this eCR to fail? Do we want this reference to be more of a "soft" foreign key vs a true constraint? If we let in random codes, only admins will be able to see them ever. Thoughts?

mcmcgrath13: How can we test that this fhir path works the way we want? All of the tests passing when this is off is concerning. Why don't any of our convert seed data fail?

mcmcgrath13: How can we test that this fhir path works the way we want? All of the tests passing when this is off is concerning. Why don't any of our convert seed data fail?

emmastephenson: Hmmm @mcmcgrath13 I think it's probably ok that only admins could see those eCRs? Would just want a way to flag that somehow. Might be worth talking over this with design to account for this edge case...maybe we want some kind of "hey we found this eCR but don't know which PA it belongs to, please review and assign manually"

mcmcgrath13: on sleeping, I'm pretty sure the fhirpath just did nothing as fhirpath is very forgiving, unit tests on the message parser could help prove out this is working as we expect maybe?

akasper: All right, I’ve made progress on all the other comments. This is the one I’m hung up on. Let’s discuss during standup today.

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/20250505100700_ecr_fks.ts
mcmcgrath13, line 1: what's with this empty file?

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/20250509000000_condition_fk.ts
mcmcgrath13: still open

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/20250509000000_condition_fk.ts
mcmcgrath13: remove foreign key per discussion at standup today

containers/ecr-viewer/src/app/api/migrate-db/migrations/core/20250509000000_condition_fk.ts
mcmcgrath13: still open

mcmcgrath13:
> I noticed that COVID-19 has two condition codes and when we backfill the condition codes we are setting condition_code in ecr_rr_conditions to whatever value came last. I'm not sure which - if either - is the correct value but I'd like to figure that out.

I was also just noticing this in the program form page. There are a few other duplicate condition names there as well. This comes from TES having multiple codes

austin-hall-skylight:
> > I noticed that COVID-19 has two condition codes and when we backfill the condition codes we are setting condition_code in ecr_rr_conditions to whatever value came last. I'm not sure which - if either - is the correct value but I'd like to figure that out.
> 
> I was also just noticing this in the program form page. There are a few other duplicate condition names there as well. This comes from TES having multiple codes
> 
> image
> I'm not sure what we should do about that in practice though...

Let's discuss at standup. Maybe normalizing the ecr_rr_condition to condition code mapping out into another table would help, but I hate to throw that wrench into things this far into it 😖

containers/ecr-viewer/integration/saveFhirDataService/core.test.ts
```diff
@@ -103,6 +139,40 @@ describe("saveFhirData - core", () => {
...
+
+    const conditions = await db
+      .selectFrom("ecr_rr_conditions")
+      .selectAll()
+      .execute();
+
+    expect(resp.message).toEqual("Success. Saved metadata to database.");
+    expect(resp.status).toEqual(200);
+    expect(conditions).toHaveLength(1);
+    expect(conditions[0].condition_code).toEqual("123");
```
mcmcgrath13, line 173: given there's not a true fk anymore, do we need the setup with the user and program areas?

containers/ecr-viewer/integration/saveFhirDataService/extended.test.ts
```diff
@@ -109,6 +134,16 @@ afterEach(() => {
 });
 
 describe("saveFhirData - extended", () => {
+  beforeEach(async () => {
+    const db = getDb<Extended>();
+    await db.insertInto("user").values(adminUser).execute();
+    await db.insertInto("program_area").values(programArea).execute();
+    await db
```
mcmcgrath13, line 141: same question here

containers/ecr-viewer/integration/saveFhirDataService/core.test.ts
```diff
@@ -18,6 +18,12 @@ const baseCoreMetadata: BundleMetadata = {
   rr: [],
   report_date: "12/20/2024",
 };
+const condition_reference = {
+  code: "123",
+  concept_name: "condition (disease)",
```
mcmcgrath13, line 23: why is this condition reference and before each needed?

mcmcgrath13 (approved):

PR #701, opened 2025-05-09
feat: Add hospital admission and discharge diagnosis data to viewer

### LINKED ISSUE #570: 3.1 Spec: Hospital Admission/Discharge Diagnosis
**NOTE**: We currently **do not have any examples of these fields** in our sample eCRs. Please refer to the 3.1 spec for example snippets.

## What needs to be done

The following fields need to be added to the eCR Viewer:
- [ ] Hospital admission diagnosis
- [ ] Hospital discharge diagnosis

Note: Only show hospital admission and discharge fields in the Viewer IF those fields are in the eICR. If there is no data, do not display.

## Why it needs to be done

We want our FHIR converter templates and the eCR Viewer to match the CDA eICR 3.1 specifications.

## Acceptance Criteria

- [ ] Work with Ashton to figure out where this should live in the eCR Viewer. Figma should be updated accordingly
- [ ] Fields should be added to the eCR Viewer
- [ ] Any relevant unit/snapshot tests should be added.

## Technical Notes

FHIR conversion is not required for this ticket; these fields are already included in the FHIR converter.

**Data field OIDs:**
- Hospital admission diagnosis: `2.16.840.1.113883.10.20.22.4.34`
- Hospital discharge diagnosis: `2.16.840.1.113883.10.20.22.4.33`

## Additional context

[eCR 3.1 Specification Analysis Spreadsheet](https://cdc.sharepoint.com/:x:/r/teams/OPHDST-IRD-DIBBs/Shared%20Documents/CDC%20DIBBs%20Project/Streamline%20Team/5%20-%20Engineering/eICR%203.1%20Specification%20Analysis.xlsx?d=w15475e2dc1d346ef98891e94a9f31f95&csf=1&web=1&e=YCPBpx)
[eCR 3.1 Specification Docs](https://www.hl7.org/implement/standards/product_brief.cfm?product_id=436#:~:text=HL7%20CDA%C2%AE%20R2%20Implementation%20Guide%3A%20Public%20Health%20Case%20Report%20%2D%20the%20Electronic%20Initial%20Case%20Report%20(eICR)%20Release%202%2C%20STU%20Release%203.1%20%2D%20US%20Realm)
Note: You can search for the data field templates via their name/OIDs in the Vol. 2 PDF. It will provide context (where the field should live), the constraints of the field (`templateID`,`classCode`, etc.), and example XML sections.

**Relevant files (FHIR converter):**
- https://github.com/CDCgov/dibbs-FHIR-Converter/blob/3e85b18d8687f4fc08ce78158c5de64ff0e48d71/data/Templates/eCR/Section/_HospitalAdmissionDiagnosis.liquid
- https://github.com/CDCgov/dibbs-FHIR-Converter/blob/3e85b18d8687f4fc08ce78158c5de64ff0e48d71/data/Templates/eCR/Section/_HospitalDischargeDiagnosis.liquid
- 

## Designs

[Figma file in dev mode](https://www.figma.com/design/zAPyOEYwNtx3f2Y47aoq0G/eCR-Viewer-(Aug-2024---Present)?node-id=10650-129453&m=dev)

gordonfarrell:
# PULL REQUEST
## Summary

Adds a new Hospital Encounter Details section to the encounter that includes hospital admission and discharge diagnoses tables

## Related Issue

Fixes #570

## Acceptance Criteria

 Fields should be added to the eCR Viewer
 Any relevant unit/snapshot tests should be added.

## Additional Information

There's a new eCR being seeded on this branch with a patient name of Dash Rendar. This eCR has synthetic hospital admin/discharge diagnosis data in it for testing.

### DIFF
--- a/containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
+++ b/containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
@@ -26,11 +26,13 @@ import {
   DataDisplay,
   DisplayDataProps,
 } from "@/app/view-data/components/DataDisplay";
+import EvaluateTable from "@/app/view-data/components/EvaluateTable";
 import { ExpandCollapseAccordion } from "@/app/view-data/components/ExpandCollapseAccordion";
 import { JsonTable } from "@/app/view-data/components/JsonTable";
 
 import {
   formatDate,
+  formatDateTime,
   formatPeriodDate,
   formatStartEndDate,
   formatStartEndDateTime,
@@ -47,6 +49,7 @@ import {
   formatPhoneNumber,
   sortByPeriod,
   formatCurrentAddress,
+  formatReference,
 } from "./formatService";
 import { HtmlTableJsonRow } from "./htmlTableService";
 import { evaluateTravelHistoryTable } from "./socialHistoryService";
@@ -602,6 +605,60 @@ export const evaluateEncounterData = (fhirBundle: Bundle) => {
     },
   ];
   return evaluateData(encounterData);
+};
+
+/**
+ * Evaluates Hospital Encounter Admission and Discharge diagnosis data from the FHIR bundle and formats it into structured data for display.
+ * @param fhirBundle - The FHIR bundle containing hospital encounter data.
+ * @returns An array of evaluated and formatted hospital encounter data.
+ */
+export const evaluateHospitalEncounterData = (fhirBundle: Bundle) => {
+  const hospitalEncounterData = [
+    {
+      title: "Hospital Admission Diagnosis",
+      value: evaluateEncounterDiagnosisData(fhirBundle, "46241-6", "Hospital Admission Diagnosis"),
+      table: true
+    },
+    {
+      title: "Hospital Discharge Diagnosis",
+      value: evaluateEncounterDiagnosisData(fhirBundle, "11535-2", "Hospital Discharge Diagnosis"),
+      table: true
+    },
+  ];
+  return evaluateData(hospitalEncounterData);
+};
+
+/**
+ * Evaluates either Hospital Admission Diagnosis or Discharge diagnosis data from the FHIR bundle and formats it into structured data for display.
+ * @param fhirBundle - The FHIR bundle containing hospital encounter data.
+ * @param code - the associated code for the section to be evaluated. e.g. 46241-6 for Admission Dx or 11535-2 for Discharge Dx
+ * @param caption - A string to set the caption for the UI element
+ * @returns An array of evaluated and formatted hospital encounter data.
+ */
+export const evaluateEncounterDiagnosisData = (fhirBundle: Bundle, code: string, caption: string) => {
+
+  const dxRefs = evaluateAll(fhirBundle, fhirPathMappings.hospitalEncounterDiagnosisRef, {code})
+
+  const dx: Condition[] = dxRefs.map((x)=> {
+    return evaluateReference<Condition>(fhirBundle, formatReference(x))
+  }).filter((x):x is Condition => Boolean(x))
+
+  if(dx.length === 0) return;
+
+
+  const dxColumns = [
+    {
+      columnName: "Problem",
+      infoPath: "hospitalEncounterDiagnosisCode",
+    },
+    {
+      columnName: "Date/Time",
+      infoPath: "hospitalEncounterDiagnosisDateTime",
+      applyToValue: formatDateTime
+    },
+  ];
+
+  return <EvaluateTable resources={dx} columns={dxColumns} caption={caption}/>;
 };
 
 /**
--- a/containers/ecr-viewer/src/app/services/formatService.ts
+++ b/containers/ecr-viewer/src/app/services/formatService.ts
@@ -9,6 +9,7 @@ import {
   Period,
   Quantity,
   Range,
+  Reference,
   RelatedPerson,
 } from "fhir/r4";
 
@@ -423,6 +424,15 @@ export const formatAge = (age: Age | undefined): string | undefined => {
 
   return `${getFormattedMonths(totalMonths)}${days} day${makePlural(days)}`;
 };
+
+/**
+ * Returns the value of a Reference
+ * @param reference the reference being formatted
+ * @returns .reference value of the supplied reference
+ */
+export const formatReference = (reference: Reference): string => {
+  return reference?.reference || ""
+}
 
 const getFormattedMonths = (months: number): string => {
   if (months < 1) return "";
--- a/containers/ecr-viewer/src/app/tests/components/Unavailable.test.tsx
+++ b/containers/ecr-viewer/src/app/tests/components/Unavailable.test.tsx
@@ -48,6 +48,16 @@ describe("UnavailableInfo", () => {
         value: "",
       },
     ];
+    const hospitalEncounterUnavailableData = [
+      {
+        title: 'Hospital Admission Diagnosis',
+        value: ""
+      },
+      {
+        title: 'Hospital Discharge Diagnosis',
+        value: ""
+      }
+    ]
     const facilityUnavailableData = [
       {
         title: "Facility Name",
@@ -143,6 +153,7 @@ describe("UnavailableInfo", () => {
         demographicsUnavailableData={demographicsUnavailability}
         socialUnavailableData={socialUnavailability}
         encounterUnavailableData={encounterUnavailableData}
+        hospitalEncounterUnavailableData={hospitalEncounterUnavailableData}
         facilityUnavailableData={facilityUnavailableData}
         providerUnavailableData={providerUnavailableData}
         symptomsProblemsUnavailableData={[
--- a/containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
+++ b/containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
@@ -84,6 +84,9 @@ export type PathTypes = {
   encounterDiagnosis: EncounterDiagnosis;
   encounterType: string;
   encounterID: Identifier;
+  hospitalEncounterDiagnosisRef: Reference;
+  hospitalEncounterDiagnosisCode: CodeableConcept;
+  hospitalEncounterDiagnosisDateTime: String;
   facilityContact: string;
   facilityContactAddress: string;
   facilityLocation: string;
@@ -350,6 +353,21 @@ const _fhirPathMappings: { [K in FhirPathKeys]: Omit<FhirPath<K>, "name"> } = {
     type: "Identifier",
     path: "Bundle.entry.resource.where(resourceType = 'Encounter')[0].identifier",
   },
+
+  hospitalEncounterDiagnosisRef: {
+    type: "Reference",
+    path: "Bundle.entry.resource.where(resourceType = 'Composition').section.where(code.coding.where(code = %code).exists()).entry"
+  },
+
+  hospitalEncounterDiagnosisCode:{
+    type: "CodeableConcept",
+    path: "Condition.code"
+  },
+  hospitalEncounterDiagnosisDateTime:{
+    type: "string",
+    path: "Condition.onsetDateTime"
+  },
+
   facilityContact: {
     type: "string",
     path: "Bundle.entry.resource.where(resourceType = 'Location')[0].telecom.where(system = 'phone')[0].value",
@@ -374,11 +392,6 @@ const _fhirPathMappings: { [K in FhirPathKeys]: Omit<FhirPath<K>, "name"> } = {
     type: "ValueX",
     path: "Bundle.entry.resource.where(resourceType = 'Encounter')[0].location[0].extension.where(url = 'http://build.fhir.org/ig/HL7/case-reporting/StructureDefinition-us-ph-location-definitions.html//Location.type').value",
   },
-
-  // hospitalAdmissionDiagnosis: {
-  //
-  // }
-
   compositionEncounterRef: {
     type: "string",
     path: "Bundle.entry.resource.where(resourceType = 'Composition').encounter.reference",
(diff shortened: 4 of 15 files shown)


containers/ecr-viewer/src/app/tests/components/Unavailable.test.tsx
```diff
@@ -48,6 +48,16 @@ describe("UnavailableInfo", () => {
         value: "",
       },
     ];
+    const hospitalEncounterUnavailableData = [
+      {
+        title: "Hospital Admission Diagnosis",
+        value: "",
```
mcmcgrath13, line 54: I know this is following a pattern, but it would probably be a better test to actually do `evaluateHospitalEncounterData([])` and get back what the tru empty case looks like instead of hard coding here (not sure if `[]` will pass as a bundle, might need to be an object? or minimally defined `Bundle` (there are some of these kicking around elsewhere in the tests))

containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
```diff
@@ -350,6 +353,21 @@ const _fhirPathMappings: { [K in FhirPathKeys]: Omit<FhirPath<K>, "name"> } = {
     type: "Identifier",
     path: "Bundle.entry.resource.where(resourceType = 'Encounter')[0].identifier",
   },
+
+  hospitalEncounterDiagnosisRef: {
+    type: "Reference",
+    path: "Bundle.entry.resource.where(resourceType = 'Composition').section.where(code.coding.where(code = %code).exists()).entry",
```
mcmcgrath13, line 359:
```suggestion
    path: "Bundle.entry.resource.where(resourceType = 'Composition').section.where(code.coding.code = %code).entry",
```

is this equivalent? the exists threw me a bit on readability

containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
```diff
@@ -467,7 +484,7 @@ const _fhirPathMappings: { [K in FhirPathKeys]: Omit<FhirPath<K>, "name"> } = {
     path: "Bundle.entry.resource.section.where(code.coding[0].code = '29549-3').entry.reference",
   },
 
-  // CareTEam
+  // CareTeam
```
mcmcgrath13, line 487: 🧹 ❤️

containers/ecr-viewer/src/app/tests/components/Unavailable.test.tsx
gordonfarrell: Ahhhh good point, let me see what it does

containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
gordonfarrell: It is equivalent, swapped to that. Thanks!

containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
```diff
@@ -84,6 +84,9 @@ export type PathTypes = {
   encounterDiagnosis: EncounterDiagnosis;
   encounterType: string;
   encounterID: Identifier;
+  hospitalEncounterDiagnosisRef: Reference;
+  hospitalEncounterDiagnosisCode: CodeableConcept;
+  hospitalEncounterDiagnosisDateTime: String;
```
mcmcgrath13, line 89:
```suggestion
  hospitalEncounterDiagnosisDateTime: string;
```

containers/ecr-viewer/src/app/services/formatService.ts
```diff
@@ -424,6 +425,15 @@ export const formatAge = (age: Age | undefined): string | undefined => {
...
 };
 
+/**
+ * Returns the value of a Reference
+ * @param reference the reference being formatted
+ * @returns .reference value of the supplied reference
+ */
+export const formatReference = (reference: Reference): string => {
+  return reference?.reference || "";
+};
```
mcmcgrath13, line 435:
```suggestion
export const formatReference = (reference: Reference | undefined): string | undefined => {
  if (!reference) return;

  return reference.reference;
};
```

This means we'll need the `|| ""` at the call site for `evaluateValue`, but it's more consistent with how most of these formatters work. Can you also move this to above `formatAge` - having this sandwiched between it and its helper below is a bit odd in terms of file flow

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
```diff
@@ -604,6 +607,76 @@ export const evaluateEncounterData = (fhirBundle: Bundle) => {
...
+};
+
+/**
+ * Evaluates either Hospital Admission Diagnosis or Discharge diagnosis data from the FHIR bundle and formats it into structured data for display.
+ * @param fhirBundle - The FHIR bundle containing hospital encounter data.
+ * @param code - the associated code for the section to be evaluated. e.g. 46241-6 for Admission Dx or 11535-2 for Discharge Dx
+ * @param caption - A string to set the caption for the UI element
+ * @returns An array of evaluated and formatted hospital encounter data.
+ */
+export const evaluateEncounterDiagnosisData = (
```
mcmcgrath13, line 646: could you add a test that covers this function?

containers/ecr-viewer/src/app/services/formatService.ts
```diff
@@ -424,6 +425,15 @@ export const formatAge = (age: Age | undefined): string | undefined => {
   return `${getFormattedMonths(totalMonths)}${days} day${makePlural(days)}`;
 };
 
+/**
+ * Returns the value of a Reference
+ * @param reference the reference being formatted
+ * @returns .reference value of the supplied reference
+ */
+export const formatReference = (reference: Reference): string => {
```
JNygaard-Skylight, line 433: I'm not sure if I understand the reason for this function. It's just returning the value of `reference`, and not doing any formatting to the value. Perhaps this should be renamed/moved?

containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
```diff
@@ -352,6 +355,21 @@ const _fhirPathMappings: { [K in FhirPathKeys]: Omit<FhirPath<K>, "name"> } = {
...
   },
+
+  hospitalEncounterDiagnosisRef: {
+    type: "Reference",
+    path: "Bundle.entry.resource.where(resourceType = 'Composition').section.where(code.coding.code = %code).entry",
+  },
+
+  hospitalEncounterDiagnosisCode: {
+    type: "CodeableConcept",
+    path: "Condition.code",
```
JNygaard-Skylight, line 366: We have other paths like this, so for consistency I think this is okay.  But just to see what you think: this could be used for anything with `Condition.code`, perhaps it should have a more generic name?

containers/ecr-viewer/src/app/services/formatService.ts
gordonfarrell: At this point it doesn't really do anything, it's basically just a skeleton to be filled in later with more logic. Mary M requested it be popped out into a function like this.

mcmcgrath13: 🙋 Yeah, I'm just doing some light future proofing for when we finally see a more complex reference and need to deal with it. It might be YAGNI, but it's also following an established pattern we have, so I can live with that risk

mcmcgrath13: @gordonfarrell could you maybe add a comment to that effect?

containers/ecr-viewer/src/app/services/formatService.ts
gordonfarrell: Swapped this around as you mentioned, let me know how it looks

containers/ecr-viewer/src/app/services/formatService.ts
```diff
@@ -402,6 +403,18 @@ export const formatRange = (data: Range | undefined): string | undefined => {
   }
 };
 
+/**
+ * Returns the value of a Reference
```
mcmcgrath13, line 407:
```suggestion
 * Returns the value of a Reference. While this function is currently very simple, it exists to future proof a change in how we format references.
```

containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
mcmcgrath13: This is also something I've been contemplating. Maybe a subset of "helper" paths vs specific paths or something? (doesn't need to be this PR, not sure if this would really improve readability/maintainability either)

containers/ecr-viewer/src/app/tests/services/evaluateFhirDataServices.test.ts
```diff
@@ -427,6 +430,42 @@ Home: 123-456-6909`,
...
+      const actual = evaluateHospitalEncounterData(
+          BundleHospitalEncounterOnlyAdmissionDx as unknown as Bundle,
+      );
+
+      expect(actual).toMatchSnapshot();
+    });
+
+    it("A bundle with only Discharge Diagnosis returns that data and matches snapshot", () => {
+      const actual = evaluateHospitalEncounterData(
+          BundleHospitalEncounterOnlyDischargeDx as unknown as Bundle,
```
mcmcgrath13: instead of having three new bundles here, can we use one and add/remove the admission/discharge references as needed? (there are some examples of this elsewhere - maybe labs?)

containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
gordonfarrell: Renamed this and the one below to "ConditionCode" and "ConditionOnsetDateTime" to be more generic

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
gordonfarrell: Tests added!

containers/ecr-viewer/src/app/tests/services/evaluateFhirDataServices.test.ts
mcmcgrath13: could we also add some sanity check that actual is what we want besides the snapshot? (I can see someone (perhaps, me 😬 ) blindly updating snapshots at some point and missing something)

gordonfarrell: I'll take a look

containers/ecr-viewer/src/app/services/formatService.ts
gordonfarrell: Added

containers/ecr-viewer/src/app/tests/services/evaluateFhirDataServices.test.ts
gordonfarrell: Added additional checks and also a somewhat funky system to push necessary admission and discharge dx data into a fhir bundle for testing

containers/ecr-viewer/src/app/services/formatService.ts
gordonfarrell: Comment is added

containers/ecr-viewer/src/app/tests/services/evaluateFhirDataServices.test.ts
```diff
@@ -456,25 +504,59 @@ Home: 123-456-6909`,
...
+      const bundle = JSON.parse(JSON.stringify(BundlePatientWithCovid)) as Bundle;
+
+      // Grab the composition section
+      const compositionEntry = bundle!.entry!.find(
+          (e: any) => e.resource?.resourceType === "Composition"
+      );
+
+      // Add Hospital encounter data to section array. ts-ignore because it thinks section doesn't exist
+      // @ts-ignore
+      compositionEntry?.resource?.section?.push(admissionDiagnosis, dischargeDiagnosis)
```
mcmcgrath13: can we import bundle once and do something like this instead

```
const thisBundle = {
   ...bundle,
   entry: [
      { 
         ...bundle.entry[0],
         section: [
            ...bundle.entry[0].section,
            dxSection
         ]
      },
      ...bundle.entry.slice(1)
   ]
}
```

it's a bit clunky, but is non mutating and could probably be abstracted out to a function so you could reuse for the various tests

containers/ecr-viewer/src/app/tests/services/evaluateFhirDataServices.test.ts
```diff
@@ -427,6 +435,164 @@ Home: 123-456-6909`,
...
+
+      const actual = evaluateHospitalEncounterData(
+        bundleWithHospitalEncounterData,
+      );
+
+      expect(actual).toMatchSnapshot();
+      expect(actual.availableData[0].title).toEqual(
+        "Hospital Admission Diagnosis",
+      );
+      expect(actual.availableData[1].title).toEqual(
```
mcmcgrath13, line 548: can we check the diagnosis text here as well? make sure we're pulling out the code/datetime correctly

containers/ecr-viewer/src/app/services/evaluateFhirDataService.tsx
```diff
@@ -604,6 +607,76 @@ export const evaluateEncounterData = (fhirBundle: Bundle) => {
...
+  if (dx.length === 0) return;
+
+  const dxColumns = [
+    {
+      columnName: "Problem",
+      infoPath: "hospitalEncounterDiagnosisCode",
+    },
+    {
+      columnName: "Date/Time",
+      infoPath: "hospitalEncounterDiagnosisDateTime",
```
mcmcgrath13, line 672: these info paths need to be updated to the new names (and we should make sure tests catch that this scenario)

mcmcgrath13 (approved):
thanks!

PR #712, opened 2025-05-12
feat: `process-ecr` api endpoint

### LINKED ISSUE #698: Add process-message route to viewer
## What needs to be done

Create a new `/process-message` endpoint on the eCR Viewer, to replace the one that exists today in Orchestration.

**Please keep the method signature / parameter names of the existing Orchestration endpoint.** This endpoint will be used directly by our users as they call the pipeline, so it's important not to change this too drastically for them. Some params may not longer be necessary, but for the ones that are shared, please keep the names the same.

## Why it needs to be done

We want this endpoint to be protected by the viewer's authentication.

## Acceptance Criteria

- New `/process-message` endpoint exists on eCR Viewer
- New endpoint should:
    - Accept eICR and RR in the request body, like orchestration does today
    - Determine which orchestration configuration is applicable based on eCR Viewer environment variables (AWS, Azure, integrated, non-integrated, etc)
    - Send the eCR and RR to the orchestration service for processing
    - Accept the orchestration response
    - Save eCR data from orchestration (to S3, SQL Server, etc)

mcmcgrath13:
# PULL REQUEST

## Summary

Add the `process-ecr` endpoint to the viewer. Similar to the `process-zip` endpoint for the viewer, hardcode the fields that are known and only ask for the data. The message endpoint supports both form and json data (because Rhapsody 6 can't do forms in a sane way 😭 ).

Rewrite `process-zip` requests into `process-ecr` via middleware and have the logic for getting the data and putting it in the right place over on `process-ecr`. Left a readme to that effect in the api folder so that lost souls might find their way.

## Related Issue

Fixes #698

## Acceptance Criteria

New /process-message endpoint exists on eCR Viewer
New endpoint should:
Accept eICR and RR in the request body, like orchestration does today
Determine which orchestration configuration is applicable based on eCR Viewer environment variables (AWS, Azure, integrated, non-integrated, etc)
Send the eCR and RR to the orchestration service for processing
Accept the orchestration response
Save eCR data from orchestration (to S3, SQL Server, etc)


containers/ecr-viewer/guide.md
```diff
@@ -141,11 +141,19 @@ Rhapsody documentation and an example route can be found [here](https://github.c
 
 ### From API
 
-Data can be added directly via API requeset to eCR Viewer's `/process-zip` endpoint. See the [API documentation](./api-documentation.md) for more details.
+Data can be added directly via API requeset to eCR Viewer's `/process-ecr` endpoints. See the [API documentation](./api-documentation.md) for more details.
```
mcmcgrath13, line 144:
```suggestion
Data can be added directly via API requeset to eCR Viewer's `/process-ecr` endpoint. See the [API documentation](./api-documentation.md) for more details.
```

JNygaard-Skylight (approved):
LGTM!

PR #725, opened 2025-05-16
fix: update rr and eicr combine function for 3.1 spec

### LINKED ISSUE #723: eICR & RR won't combine for certain 3.1 eCRs
## Description

For the "Eve Everywoman" 3.1 eCR, the RR and eICR won't combine during the FHIR converter service. This means we don't convert the eICR Processing Status & any other eRSD warnings/information.

## Expected Result

Relevant RR fields (ex. eICR processing status section) should get added to the eICR.

angelathe:
# PULL REQUEST

## Summary

For the "Eve Everywoman" 3.1 eCR, the RR and eICR won't combine during the FHIR converter service. This means we don't convert the eICR Processing Status & any other eRSD warnings/information.

Fix: Starting in v3, eICRs can optionally include an RR section that duplicates content from the actual RR. In the `add_rr_to_eicr` function, we were checking for an existing merge by looking for the RR LOINC code. However, with 3.1 eICRs, this check was incorrectly returning `true` because of the new optional RR section. We updated the logic to check for the `eICR Processing Status` field instead, which is required and only present in the actual RR.
- If optional RR section exists from >3.1 eICR, we remove that section to avoid duplicate info
- Also added tests & 3.1 data for testing
- Updated eICR Version display

## Related Issue

Fixes #723 

## Acceptance Criteria

- [ ] Relevant RR fields (ex. eICR processing status section) should get added to the eICR.

## Additional Information

## Checklist


containers/fhir-converter/app/main.py
```diff
@@ -94,6 +94,21 @@ async def convert(
     return result
 
 
+@app.post("/combine")
```
angelathe, line 97: TODO ANGELA: Remove before merge

containers/ecr-viewer/src/app/services/ecrMetadataService.ts
```diff
@@ -115,7 +115,9 @@ export const evaluateEcrMetadata = (fhirBundle: Bundle): EcrMetadata => {
     if (releaseVersion === "2016-12-01") {
       return "R1.1 (2016-12-01)";
     } else if (releaseVersion === "2021-01-01") {
-      return "R3.1 (2021-01-01)";
+      return "R3 (2021-01-01)";
+    } else if (releaseVersion === "2022-05-01") {
+      return "R3.1 (2022-05-01)";
     } else {
       return releaseVersion;
     }
```
mcmcgrath13, line 123: if we think this will get any bigger it might be worth making a little mapping object from release version to name and do something like `return releaseVersionMap[releaseVersion] || releaseVersion`

containers/fhir-converter/app/main.py
```diff
@@ -206,6 +207,23 @@ def add_rr_data_to_eicr(rr, ecr):
         # Append the ecr section into the eCR - puts it at the end
         ecr.append(ecr_section)
 
+    # If eICR >=R3, remove (optional) RR section that came from eICR
+    # This is duplicate/incomplete info from RR
```
mcmcgrath13, line 211: out of curiosity, what happens if we don't remove it?

containers/fhir-converter/app/main.py
```diff
@@ -206,6 +207,23 @@ def add_rr_data_to_eicr(rr, ecr):
...
+    ecr_version = ecr.xpath('//*[@root="2.16.840.1.113883.10.20.15.2"]/@extension')
+    if (len(ecr_version) > 0) and (
+        ecr_version[0] == "2021-01-01" or ecr_version[0] == "2022-05-01"
+    ):
+        namespaces = {"hl7": "urn:hl7-org:v3"}
+        rr_from_eicr = ecr.xpath(
+            '//hl7:component[hl7:section/hl7:templateId[@root="2.16.840.1.113883.10.20.15.2.2.5" and @extension="2021-01-01"]]',
+            namespaces=namespaces,
```
mcmcgrath13, line 219: should we remove this rr info before adding in the sepearate rr? I don't know if this is founded, but it feels slightly safer to remove things before we add things just to be extra sure we don't remove info from the rr

containers/ecr-viewer/src/app/services/ecrMetadataService.ts
angelathe: Added it now, it's way cleaner

containers/fhir-converter/app/main.py
```diff
@@ -206,6 +207,23 @@ def add_rr_data_to_eicr(rr, ecr):
         # Append the ecr section into the eCR - puts it at the end
         ecr.append(ecr_section)
 
+    # If eICR >=R3, remove (optional) RR section that came from eICR
+    # This is duplicate/incomplete info from RR
+    ecr_version = ecr.xpath('//*[@root="2.16.840.1.113883.10.20.15.2"]/@extension')
+    if (len(ecr_version) > 0) and (
+        ecr_version[0] == "2021-01-01" or ecr_version[0] == "2022-05-01"
+    ):
```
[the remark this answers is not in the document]
angelathe: Ah, good catch

mcmcgrath13 (approved):
⛵

PR #769, opened 2025-05-27
fix: Remove `cache` wrapper on `getLoggedInUser` for now

mcmcgrath13:
# PULL REQUEST

## Summary

We're seeing a mystery bug where a user is somehow logged in, but their UUID/User is unavailable. @JNygaard-Skylight first saw this locally and we chalked it up to a strange dev state since logging out and logging back in fixed it, but we're now seeing it on the deployed version of the app. 

The leading culprit is the `cache` wrapping of the `getLoggedInUser` function. This is experimental in react@18, but is stable in react@19 - perhaps we try again later


JNygaard-Skylight (approved):
Approved@

PR #770, opened 2025-05-27
fix: explicit error handling and returning for actions

### LINKED ISSUE #756: Form error messages - fix on built app

mcmcgrath13:
# PULL REQUEST

## Summary

Make actions a separate concept from the base functions that do the action. Add the `errorService` which provides a new error class `UserFacingError`, which indicates that the error message is intended for users and should be passed along instead of a generic `action failed` message. Add a wrapper function to turn regular functions into actions.

## Related Issue

Fixes #756


containers/ecr-viewer/src/app/tests/services/errorService.test.ts
```diff
@@ -0,0 +1,51 @@
+import { UserFacingError, makeServerAction } from "@/app/services/errorService";
+
+describe("error service", () => {
+  describe("makeServerAction", () => {
+    const myFn = async (cse: string) => {
+      switch (cse) {
+        case "happy":
+          return 123;
+        case "expected":
```
angelathe, line 9: tiny tiny nit, can we rename this to something like `user-error`? We use `expected` in the other tests in other contexts, not sure what expected really means here (expected user error?)

angelathe (approved):

PR #782, opened 2025-05-30
fix: handling duplicate condition names

### LINKED ISSUE #773: Update condition reference table: Get condition groups
Our current `condition_reference` table is at the `condition` level, but we need to instead think about permissions in terms of Condition Groups. Where RCKMS Category (e.g. Respiratory Diseases) > Condition Group (COVID-19) > Condition (COVID-19 (but backed by a snomed concept code such as COVID-19 Infection)) > Concept (Cough).

The condition groups are maintained in TES and we need to add this level of the heirarchy to our `tes.db`. We should add the condition group ID to the `condition` table. This can be done by using the below query to pull out a map from condition value set ids to condition group value set ids then add the id to the `Condition` model in `seed_tes_data.py`

API Query to get the condition groupers:
https://tes.tools.aimsplatform.org/api/fhir/ValueSet?context=http://aphl.org/fhir/vsm/CodeSystem/usage-context-type%7Ccondition-grouper

## Acceptance Criteria

* `condition` table in `tes.db` has a `group_id` column populated with the UUID from TES

mcmcgrath13:
# PULL REQUEST

## Summary

This started as a PR to go get condition group IDs, but turned into a PR to update the display of duplicately named conditions in the program area form.

It turns out that a) not all conditions have a condition group (2 are missing) and b) condition groups can contain non-name matching conditions and c) condition groups can contain conditions that span different categories. So condition groups are not the savior we thought they would be.

After chatting with @ashton-skylight we landed on adding the concept name if available and if not the snomed code for conditions that have a duplicate name.

## Related Issue

Fixes #773 

## Additional Information

image

### DIFF
--- a/containers/ecr-viewer/e2e/dual/admin-program.spec.ts
+++ b/containers/ecr-viewer/e2e/dual/admin-program.spec.ts
@@ -49,7 +49,7 @@ test.describe("program management page", () => {
     await page.getByLabel("Program area name").fill(conditionName);
 
     await page
-      .getByRole("button", { name: "Create program area" })
+      .getByRole("button", { name: "Save program area" })
       .first()
       .click();
 
@@ -96,7 +96,7 @@ test.describe("program management page", () => {
     await page.getByLabel("Program area name").fill(conditionName);
 
     await page
-      .getByRole("button", { name: "Create program area" })
+      .getByRole("button", { name: "Save program area" })
       .first()
       .click();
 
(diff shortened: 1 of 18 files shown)


akasper (approved):
Given the limitations of condition groups, this makes perfect sense: We're making a UI change instead of a data change. Implementation works on my machine, code looks good. :shipit:

PR #831, opened 2025-06-02
feat: auto sign out

### LINKED ISSUE #758: Add automatic signout to eCR Viewer
## What needs to be done

For security purposes, users of the eCR Viewer should be logged out after a certain period - we'll start with 30m of inactivity.

Only applies to non-integrated or dual boot setups.

## Why it needs to be done

This is required for security within our pilot sites.

## Acceptance Criteria

- [ ] Users are automatically logged out from any IDP used for Viewer authentication after 30 minutes 
- [ ] Users are not logged out of their IDP everywhere - just logged out of the Viewer

mcmcgrath13:
# PULL REQUEST

## Summary

* Shorten our session timeout to 30 minutes and token timeout to 16 hours. 
* If a user is active on the page, update their session to extend it.
* When a user has less than 90 seconds left on their session, show a modal warning them about this and offering to extend or sign out

Side quests:
* Disallow `initiallyOpen` on our modals, since it plays weird with aria things
* Only look in the `main` element for side nav links (was picking up the session expiring header from the modal)
* Remove the page-leave-blocking confirmation on the forms as it played badly with session timeouts and these forms aren't that big where losing state is a big deal
* Add `zIndex` prop to modal to control how that modal appears in relation to other modals if semantic ordering isn't viable for the scenario (such as for the auto sign out where the modal is at the app level and must always be the topmost modal)

## Related Issue

Fixes #758 

## Acceptance Criteria

- [ ] Users are automatically logged out from any IDP used for Viewer authentication after 30 minutes 
- [ ] Users are not logged out of their IDP everywhere - just logged out of the Viewer

## Additional Information

image


containers/ecr-viewer/src/app/components/AutoSignout.tsx
```diff
@@ -0,0 +1,126 @@
...
+  // Keep track of seconds left until the session expires
+  useEffect(() => {
+    if (data?.expires) {
+      const expires = new Date(data.expires);
+      const ttExpire = Math.floor((expires.valueOf() - Date.now()) / 1000);
+      setTimeToExpireSecs(ttExpire);
+      const t = setTimeout(signOutGoHome, ttExpire * 1000);
+
+      // decrement time each second
+      const i = setInterval(
```
gordonfarrell, line 64: Nit: What do you think about using variables names more descriptive than t, i, or even the tt part of ttExpire? My opinion isn't super strong on this, but I tend to like more descriptive names, even if its just "time" "interval" "timeToExpire" etc for readability

gordonfarrell (approved):
Looks great, thanks for the variable names! Appreciate all the comments guiding through the autoSignout logic. Tested locally and worked perfectly.

PR #855, opened 2025-06-06
fix: get user/program management actually working

### LINKED ISSUE #765: Make server actions (or similar) work in deployed setup
Because we use an application gateway in front of our containers on azure, the referring url does not match the server's URL causing the post request a server action emits to be rejected. See [slack thread](https://skylight-hq.slack.com/archives/C05B5R8K0TF/p1748029413599279?thread_ts=1747850858.258819&cid=C05B5R8K0TF) for more details

mcmcgrath13:
# PULL REQUEST

## Summary

Various fixes to get user/program management to work for real:
* If azure AD doesn't return an email, fall back to the upn if it looks email-like
* segregate the server actions to their own service, remove `"use server"` from the db-based services as it causes everything to become async even if you don't mark it as such
* add a test to make sure non-admins can't see admin things (add another keycloak user to enable this)
* make the jsdoc requirement a lint warning instead of error (because it was annoying me)
* Add back the react cache since it wasn't the culprit (apologies to react-cache)
* Make the allowed origin more specific as wildcards are only allowed at the subdomain level

## Related Issue

Fixes #765


angelathe (approved):
Works great, thank you 🎉

PR #886, opened 2025-06-16
fix: add missing dep on `sharp`

mcmcgrath13:
# PULL REQUEST

## Summary

I was seeing an error about image optimization failing because of `sharp` not being installed, so this adds it: https://nextjs.org/docs/messages/sharp-missing-in-production


mcmcgrath13:
> Do you see what the Snyk error is? I get a 404 when I try to open the test.

I also get a 404 - I think snyk is confused

PR #893, opened 2025-06-17
build: more, faster, async-ier workers

### LINKED ISSUE #891: SPIKE: Reproduce Philly Orchestration Issues
## What needs to be done

Try to recreate the load issues Philly is seeing on the orchestration service.

## Why it needs to be done

PDPH is unable to send 10 eCRs concurrently - the orchestration container begins to register as unhealthy.

## Timebox

Until it's reproduced

## To-do list

- Reproduce Philly's setup in AWS dev
- Try to send 10+ messages through the Viewer pipeline and see when the container starts to have issues

## Additional Context
PDPH is using an async library to send the requests, rather than wait for a response - we'll need to simulate this behavior.

mcmcgrath13:
# PULL REQUEST

## Summary

Various small changes to help improve performance in our python containers:
* use `uvicorn[standard]` instead of just `uvicorn` as this will use `uvloop` as a backend, which is a wee bit faster
* make sure the fhir converter supports converting multiple files at a time by adding a random id into the input/output file names (and then cleaning them up at the end)
* use `httpx` with async requests instead of synchronous `requests` in the orchestration service's outbound http calls - this should reduce any time spent blocking waiting for messages to process
* use multiple workers when converting seed data to ensure some test coverage of this setup

Other cleanups:
* Improve some error handling/logging in `process-ecr`
* Remove `$` from f string curlies in python tests
* add `f` to string in create seed data that clearly intends to interpolate

## Related Issue

Fixes #891

## Additional Information

spreadsheet with some performance testing data: https://docs.google.com/spreadsheets/d/1v1N6GEa4Uk6u7eiet7NzmHGpPaLr2jJGqWtsO0nssAc/edit?gid=0#gid=0


containers/ecr-viewer/seed-scripts/create-seed-data.py
mcmcgrath13, line 1: this whole file will get reverted before merge

containers/ecr-viewer/src/middlewares/withUnauthorized.ts
```diff
@@ -21,7 +21,7 @@ export const withUnauthorized: MiddlewareFactory = (
     if (request.nextUrl.pathname.endsWith("/save-fhir-data")) {
       if (
         request.headers.get("x-orchestration") === "true" &&
-        request.headers.get("user-agent")?.startsWith("python-requests/")
+        request.headers.get("user-agent")?.startsWith("python-httpx/")
```
JNygaard-Skylight, line 24: What does this change do?

containers/orchestration/pytest.ini
```diff
@@ -14,4 +14,4 @@ markers =
 log_cli = true
 log_cli_level = INFO
 log_cli_format = %(asctime)s [%(levelname)8s] %(message)s (%(filename)s:%(lineno)s)
-log_cli_date_format = %Y-%m-%d %H:%M:%S
\ No newline at end of file
+log_cli_date_format = %Y-%m-%d %H:%M:%S
```
JNygaard-Skylight, line 17: Why is Git showing a change here?

austin-hall-skylight (approved):

containers/orchestration/README.md
```diff
@@ -21,7 +21,7 @@ Before running the orchestration unit tests, make sure you have all the services
 2. `eval "$(pyenv init -)"`
 3. `source .venv/bin/activate`
 4. `pip install -r requirements.txt -r dev-requirements.txt`
-5. `python -m pytest --cov-report xml --cov=. -m "not integration" tests/`
+5. `python -m "not integration" tests/`
```
JNygaard-Skylight, line 24: I think it would still be useful to have information on how to run the tests with coverage. Also you broke the command.
```suggestion
5. `python -m  pytest -m "not integration" tests/`
```

containers/ecr-viewer/.env.sample
```diff
@@ -3,6 +3,10 @@ CONFIG_NAME=AWS_SQLSERVER_DUAL
 # Seed script configuration (default, "star-wars")
 # SEED_DATA_DIRECTORIES=DIR1,DIR2
 
+# optional - number of workers on the python containers - performance tuning parameter
```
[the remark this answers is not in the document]
mcmcgrath13: it's not exactly for local dev only as we do expect it to be set in production, but it's a runtime variable, so what you set here will not affect production at all and we don't personally look at it in our application anywhere. Whereas other things we mark local dev only are really for dev-only infrastructure and this feels different to me

containers/orchestration/README.md
mcmcgrath13: I got annoyed at taking the coverage piece out yet again when copy pasting for local dev. The CI uses the coverage style one, so it is there for the looking and is a pretty standard pytest flag if one ends up at google, but for the readme I think it's rare one would want the coverage locally

containers/ecr-viewer/.env.sample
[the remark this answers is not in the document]
JNygaard-Skylight: `DevOps Configuation`? `Server Configuration` I'm really not sure.

PR #896, opened 2025-06-18
fix: sticky admin form header

mcmcgrath13:
# PULL REQUEST

## Summary

Somewhere along the way, the user/program area form header lost its stickiness - bring it back.


angelathe (approved):

angelathe:
Works for me!

PR #931, opened 2025-07-02
feat: clear library search with X or Esc

### LINKED ISSUE #755: Clear search
## User Story

As an eCR Library user, I want to clear search so that I can get back to the default Library state.

## Context 
After doing some research with other systems, databases, and file mechanisms seem to clear the search bar AND results upon hitting the `X` button. This includes Notion tables and Mac Finder.

On websites where the search corresponds with looking for pages, as in Github or GSA.gov, the `X` button corresponds with ONLY clearing the keyword/phrase in the search bar. We currently utilize the USWDS search component, which follows this website page search pattern: `X` only clears the keyword in the search bar.

On the eCR Viewer, the USWDS search component functions as it was built. To reset the table results, the user must (1) click or tab on the `X` button and then (2) click or tab the search button. Alternatively, the user can remove "search=x" from the URL. This action is unclear and not easy for a user to accomplish. To maintain the same visual style, we are changing the function of the `X` button. The `X` button will clear both the keyword/phrase in the search bar and the searched table results. 

## Acceptance Criteria

- [ ] When the 'X' button is clicked or tabbed, both the keyword/phrase in the search bar and the searched table results will clear, bringing the user to the default library state.
- [ ] Using the escape key to clear the search bar should also clear the search filter

## Scenarios

- I search for a patient in the Library
- I see what I need to and want to get back to the default Library page
- I can quickly and easily clear my search

## Design Notes

[Search - Clear Figma File in dev mode](https://www.figma.com/design/zAPyOEYwNtx3f2Y47aoq0G/eCR-Viewer--Aug-2024---Present-?node-id=12405-33327&m=dev)
[Search - Clear Figma File](https://www.figma.com/design/zAPyOEYwNtx3f2Y47aoq0G/eCR-Viewer--Aug-2024---Present-?node-id=12405-33327&t=hQ1BHdyUNkumGdOS-1)
[Search - Clear Prototype](https://www.figma.com/design/zAPyOEYwNtx3f2Y47aoq0G/eCR-Viewer--Aug-2024---Present-?node-id=12421-24504&t=hQ1BHdyUNkumGdOS-1)

## Additional context

Today you can clear search by searching for an empty string, but this behavior isn't terribly intuitive - we want something better for users.

gordonfarrell:
# PULL REQUEST
## Summary

This PR changes the behavior of the eCR Library search bar, making it so that clearing the search field by clicking the small X button or hitting esc when focused on the search input clear the search params and show the entire library of eCRs again. Previously, the x or esc just cleared the input but didn't clear the search.

Due to the way USWDS has implemented their search component we don't have access to any of the inner workings of the field or X button. In fact the X is (as far as I was able to find) impossible to interact with or even inspect in the browser. My solution reflects that by being a bit work around-y and taking more control of the state of the input using a hook updated on change. A useEffect then studies the content of the input and clears the search param when the input is populated with an empty string. This means that pressing ESC in the filled search field, the X, or manually clearing the string with backspace will all return the library to an un-searched state.

## Related Issue

Fixes #755

## Acceptance Criteria

- [ ] When the 'X' button is clicked or tabbed, both the keyword/phrase in the search bar and the searched table results will clear, bringing the user to the default library state.
- [ ] Using the escape key to clear the search bar should also clear the search filter

## Additional Information

It mentions the 'X' being tabbed to in the AC, but this actually isn't even possible with USWDS's implementation


JNygaard-Skylight (approved):
LGTM!

containers/ecr-viewer/src/app/components/LibrarySearch.tsx
```diff
@@ -23,24 +25,51 @@ const LibrarySearch = ({
   className,
   textBoxClassName,
 }: LibrarySearchProps) => {
-  const { updateQueryParam, pushQueryUpdate } = useLibraryQueryParam();
+  const [searchTerm, setSearchTerm] = useState<string | undefined>();
+  const prevSearchTermRef = useRef(searchTerm);
```
mcmcgrath13, line 29: out of curiosity, why a ref here vs a state variable?

containers/ecr-viewer/tests/e2e/dual/ecr-library.spec.ts
```diff
@@ -71,6 +71,37 @@ test.describe("ecr library page", () => {
...
+      await expect(page.getByText("Showing 1-1 of 1 eCRs")).toBeVisible();
+      await expect(
+        page.getByRole("gridcell", { name: "Minch YodaV1\nDOB: 01/01/1125" }),
+      ).toBeVisible();
+      expect(
+        (await page.locator("tbody > tr").allTextContents()).length,
+      ).toEqual(1);
+
+      // This is a workaround to simulate the effect of pressing ESC on the search box which clears it to empty string.
+      // Playwright's functions for simulating key presses did not work for this test for some reason
```
mcmcgrath13, line 95: was this consistent across all broswers?

containers/ecr-viewer/src/app/components/LibrarySearch.tsx
```diff
@@ -23,24 +25,51 @@ const LibrarySearch = ({
...
+  const { updateQueryParam, pushQueryUpdate, deleteQueryParam } =
+    useLibraryQueryParam();
+
+  useEffect(() => {
+    setSearchTerm(initSearchTerm);
+  }, [initSearchTerm]);
```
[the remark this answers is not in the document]
gordonfarrell: I did this because I was seeing some issues where the value of the search input would be lost on refresh unless hydrated in the useEffect. I switched it back to setting it in the hook and I'm not seeing that issue anymore so maybe it was interacting weirdly in a previous version of the implementation and is good now.

Will open a tweak PR in a bit after I look at all of these

containers/ecr-viewer/src/app/components/LibrarySearch.tsx
mcmcgrath13: I don't feel strongly here either way - genuine curiosity as it wasn't what I would have first attempted and I like to learn from other folks' PRs

containers/ecr-viewer/src/app/components/LibrarySearch.tsx
```diff
@@ -23,24 +25,51 @@ const LibrarySearch = ({
...
-        e.preventDefault();
-        const searchTerm = (
-          e.currentTarget.elements.namedItem("search-field") as HTMLInputElement
-        )?.value;
-        updateQueryParam("search", searchTerm);
-        pushQueryUpdate();
-      }}
-      defaultValue={initSearchTerm}
+      onSubmit={handleSubmit}
+      onChange={(e) => setSearchTerm((e.target as HTMLInputElement).value)}
```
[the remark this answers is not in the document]
gordonfarrell: I landed on this because it's tough to interact with the inner workings of the USWDS search component. The input inside can't really be directly accessed and its change events end up as form events in the wrapping form which causes TS to get upset when I try and call .value. I couldn't get typescript to be happy without a cast, though maybe you have some ideas for what could make it work that could make this a bit nicer

PR #942, opened 2025-07-07
feat(github-actions): update AWS dev deploy workflow

alismx:
## Changes Proposed

- Updated the GitHub workflow for triggering AWS Dev deployments.
- Removed the inputs and internal references for several variables. (dibbs-tf-envs handles this now, keeps it simple and means we don't have to always know what's going on in the other repo)
- Revised the event_type options to a single specific value.
- Modified variable references to use github.event.inputs for container_tag, tf_branch, and event_type.
- Updated to use the DIBBS_SKYLIGHT_PAT like the dibbs-vm trigger workflow.

## Additional Information

- The original workflow intended for dual deployments (integrated and non-integrated) has been simplified to a single deployment path, dibbs-tf-envs will deploy the correct config to each environment.

## Testing

- Trigger the workflow_dispatch event with valid inputs (run workflow, select this branch `alis/trigger-update`, put `6.0.0` as the container tag, deploy! https://github.com/CDCgov/dibbs-ecr-viewer/actions/workflows/trigger-dev-workflow.yaml
- Verify that the workflow sends a dispatch to the dibbs-tf-envs using the updated parameters; it should deploy the correct version to dev7 and dev42. https://github.com/skylight-hq/dibbs-tf-envs/actions/workflows/deploy_dibbs_services_aws_dev.yaml
- Confirm this dibbs-tf-envs deploys to dev42 and dev7 correctly.

### DIFF
--- a/.github/workflows/trigger-dev-workflow.yaml
+++ b/.github/workflows/trigger-dev-workflow.yaml
@@ -8,20 +8,6 @@ on:
         type: string
         description: "The container tag to deploy. Example: `5.0.0`"
         required: true
-      dev_id:
-        type: choice
-        description: "The unique ID for the dev environment. `42` is typically used for dual/non-integrated deployments and `7` is typically used for integrated."
-        options:
-          - "7"
-          - "42"
-      config_name:
-        type: choice
-        description: "Sets CONFIG_NAME on dibbs-ecr-viewer"
-        required: true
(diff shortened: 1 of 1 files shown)


angelathe (approved):
Followed the testing steps and everything looked good! Thank you Alis!! 🎉

PR #948, opened 2025-07-08
fix: string sorting should ignore case

### LINKED ISSUE #904: Program area table: sort by program area
## Description

When user goes to the program management table and filters the program areas, filtering should only occur alphabetically, and not off upper or lower case.

## Steps to reproduce

1. User is on the program management table 
2. User clicks on the filter arrow to the right of program area header
3. Program areas filter in alphabetical order

## Expected Result

With the filtering by alphabetical order, the table shows: 
1. apple
2. Bacon
3. hello

## Actual result

With the current pattern of filtering by upper/lower case and alphabetical, the table shows:
1. apple
2. hello 
3. Bacon

## Priority

#### Priority

## Notes

mcmcgrath13:
# PULL REQUEST

## Summary

Add a `stringSort` helper function and use it everywhere we sort strings so that case is ignored (well, so local sorting conventions are respected)

## Related Issue

Fixes #904

## Additional Information

image


gordonfarrell (approved):
Works great! Thanks Mary

PR #952, opened 2025-07-09
feat: Pregnancy Info Section

JNygaard-Skylight:
# PULL REQUEST

## Summary

Adds the "Pregnancy Info" section. Currently on three possible fields are available:
- Pregnancy status
- Postpartum status
- Postpartum effective date/period
The rest are not currently converted, or are converted incorrectly.

## Related Issue

Fixes #573 

## Acceptance Criteria

- [ ] Fields should be added to the FHIR Converter
- [ ] Work with Ashton to figure out where this should live in the eCR Viewer. Figma should be updated accordingly
- [ ] Fields should be added to the eCR Viewer~~
- [ ] Any relevant unit/snapshot tests should be added.

Designs will be worked on in a separate ticket.

## Additional Information

The various bits of data for a pregnancy come from four different observations:
- [Last Menstrual Period](https://hl7.org/fhir/us/vr-common-library/StructureDefinition-Observation-last-menstrual-period.html)
- [US Public Health Pregnancy Status Observation
](https://hl7.org/fhir/us/ecr/2.1.2/StructureDefinition-us-ph-pregnancy-status-observation.html)
- [US Public Health Pregnancy Outcome Observation](https://hl7.org/fhir/us/ecr/2.1.2/StructureDefinition-us-ph-pregnancy-outcome-observation.html)
- [US Public Health Postpartum Status](https://hl7.org/fhir/us/ecr/2.1.2/StructureDefinition-us-ph-postpartum-status.html)

Apart from pregnancy outcome to pregnancy status, there is no explicit link between the different observations. Therefore as a temporary work-around instead of grouping by pregnancy, I am grouping by observation, in chronological order. A more satisfactory solution will be determined later.

image

## Checklist

- [ ] If this code affects the other scrum team, have they been notified? (In Slack, as reviewers, etc.)

PR #992, opened 2025-07-17
refactor: Sort observations by date

JNygaard-Skylight:
# PULL REQUEST

## Summary

changed the previous `sortByPeriod` function to be a more versatile function that can sort any observations. This function will also be used in #952 

Also created a `fhir-data-utils.ts`. The idea is most of the code in `evaluateFhirDataService` can be moved into here.


containers/ecr-viewer/src/app/view-data/utils/fhir-data-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+  return effectiveDiff || compareDates(datesA.start, datesB.start);
+};
+
+/**
+ * Return an descending order (most recent first) list of observations by `effective[x]` (period or DateTime).
+ * @param observationArray Array of observations
+ * @returns Ordered list of observations by effective[x]
+ */
+export const sortObservationsByDate = (observationArray: Observation[]) =>
+  observationArray.sort(compareObservationsByDate);
```
mcmcgrath13, line 51: could we make this more generic to `Resource<that has effictiveX>`? We also sort things like addresses and it would be good to be able to leverage that here

JNygaard-Skylight: If this worked the same way as `sortbyPeriod`, where you had to pass in a function that would be used to get the relevant date value? So for observations it would be `obs => evaluteValue(obs, FhirPathMappings.effecticeX)` and for adresses ` addr => addr.period`? That would work if they are all the same type, or do you think this should be able to work with heterogenous collections of objects?

containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
```diff
@@ -699,9 +695,13 @@ const _fhirPathMappings: { [K in FhirPathKeys]: Omit<FhirPath<K>, "name"> } = {
...
     path: "value",
   },
+  occurrenceX: {
+    type: "TimeX",
+    path: "occurrence",
+  },
 };
 
-type FhirPathMappings = { [K in FhirPathKeys]: FhirPath<K> };
+export type FhirPathMappings = { [K in FhirPathKeys]: FhirPath<K> };
```
mcmcgrath13, line 704: I'm not seeing this used anywhere else - do we need the export?

containers/ecr-viewer/src/app/view-data/utils/fhir-data-utils.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import { evaluateOne } from "@/app/utils/evaluate";
+import { FhirPath, PathTypes } from "@/app/utils/evaluate/fhir-paths";
+
+const ONGOING_DATE = new Date("9999-01-01");
+
+const parseDate = (dateString: string | undefined): Date | undefined =>
+  dateString ? new Date(dateString) : undefined;
+
+const getDates = <K extends keyof PathTypes>(
```
mcmcgrath13, line 11:
```suggestion
const getDates = <K extends FhirPathKeys>(
```

containers/ecr-viewer/src/app/view-data/utils/fhir-data-utils.ts
```diff
@@ -0,0 +1,72 @@
...
+  obs: FhirResource,
```
mcmcgrath13, line 12: given this could be a non-observation, can we change the variable name to something like `resource`?

containers/ecr-viewer/src/app/view-data/utils/fhir-data-utils.ts
```diff
@@ -0,0 +1,72 @@
...
+  datePath: FhirPath<K>,
+) => {
+  const dateElement = evaluateOne(obs, datePath) as Period | string;
```
mcmcgrath13, line 15: I think you could instead limit `FhirPathKeys` to those keys that result in a `Period | string` so that this is inherent to the function - I just did something like that with the `evaluateOneReference`

mcmcgrath13 (approved):
⚜️

PR #1003, opened 2025-07-18
fix: display unavailable data uniformly

### LINKED ISSUE #951: Bug: Full width content affecting Unavailable Info view
## Description

Noticed that fields that utilize the `fullWidthContent` option in their DisplayDataProps also have this option applied when they show up in the Unavailable Info Section.

For example, see **Occupation History** and **Disability Status**. The _No data_ content should be to the right of the field name, not below.

## Steps to reproduce

View the Dash Rendar eCR to see the Occupation History and Disability Status fields in the Unavailable Info section.

View the Eve Everywoman eCR to see these fields with available data.

## Expected Result

For fields that use the `fullWidthContent`:
1. When there is available data, the `fullWidthContent` option should be respected (done!)
2. When there is NOT available data, the formatting should follow the rest of the Unavailable Info section (_No data_ should appear to the right of the field name, not below).

Relevant Fields (any fields that use `fullWidthContent`): 
- Occupation History
- Disability Status

## Actual result

See above

## Priority

#### Priority

## Notes

mcmcgrath13:
# PULL REQUEST

## Summary

Only bring over the `title` and `toolTip` props when displaying an unavailable item. This avoids strange full width handling (and future proofs us against other such shenanigans) 

## Related Issue

Fixes #951

## Additional Information

image

### DIFF
--- a/containers/ecr-viewer/src/app/view-data/components/UnavailableInfo.tsx
+++ b/containers/ecr-viewer/src/app/view-data/components/UnavailableInfo.tsx
@@ -118,9 +118,9 @@ const UnavailableSection = ({
   return (
     data.length > 0 && (
       <AccordionSubSection title={title} idPrefix="unavailable-">
-        {data.map((item, index) => (
+        {data.map(({ title, toolTip }, index) => (
           <DataDisplay
-            item={{ ...item, value: "No data" }}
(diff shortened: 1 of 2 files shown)


angelathe (approved):
Thank you for fixing this!!

PR #1007, opened 2025-07-21
ci: bail out if azure ad + dependabot

mcmcgrath13:
# PULL REQUEST

## Summary

Dependabot doesn't have access to secrets and using azure AD in our e2e tests requires secrets. Have this still fail, but in a more obvious way.


gordonfarrell (approved):

.github/workflows/container-ecr-viewer.yaml
```diff
@@ -104,6 +104,14 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Bail out if dependabot and azure ad
```
gordonfarrell, line 107:
```suggestion
      - name: Bail out if Dependabot and Azure AD
```

Nits

PR #1016, opened 2025-07-22
feat: add audit logging to user signin/out

### LINKED ISSUE #959: Audit logging: Sign In/Out
## What needs to be done

Add audit logging support for user sign in and out

## Why it needs to be done

We need to have track of all user accesses in the log

## Acceptance Criteria

- [ ] When a user signs in, a `user` `signin` audit log is also emitted
- [ ] When a user signs out, a `user` `signout` audit log is also emitted

## Technical Notes

- Use the `audit` wrapper in the `auditLogService.ts` (`Action` type may need to be adjusted to add sign in/out)
- This will likely want to be implemented as callbacks on the next auth config

mcmcgrath13:
# PULL REQUEST

## Summary

Add audit logging to sign in and out actions. This is done via a callback for `signin` and an event listener for `signout`. I moved the updating of log in time and user name into the sign in callback for a) consistency between that datetime and audit log and b) so we call it less often. 

Other notes:
* The `loggedInUserService` was added as otherwise there was a funky circular dependency that webpack was choking on.
* Because the login date and name udpates moved to sign in, the functional testing of this moved to the e2e tests
* Typically, we test these audit functions via integration test, but given the reliance on next auth flows, I felt this could be best tested via e2e

## Related Issue

Fixes #959

## Acceptance Criteria

- [ ] When a user signs in, a `user` `signin` audit log is also emitted
- [ ] When a user signs out, a `user` `signout` audit log is also emitted


gordonfarrell (approved):
Works perfectly thanks Mary!

PR #1018, opened 2025-07-22
feat: add audit logging on ecr creation

### LINKED ISSUE #961: Copy of Audit logging: Create/Edit/Delete program area
## What needs to be done

Add audit logging support for program area CRUD

## Why it needs to be done

We need to have track of all user status/roles changes in the log, which are influenced by what program areas they have access to and what those program area has access to

## Acceptance Criteria

- [ ] When a program area is deleted a `program` `delete` audit log is also emitted
- [ ] When a program area is created a `program` `create` audit log is also emitted
- [ ] When a program area is updated a `program` `update` audit log is also emitted

## Technical Notes

mcmcgrath13:
# PULL REQUEST

## Summary

Add audit logging when eCR is created

Notes:
* Did this in a more hands on way due to all the stuff already happening with transactions and params in the saving fhir data function
* Made transaction an optional param to `getLoggedInUser` and `getUserByEmail` so they could be used more places
* This is currently into #1016 as they both do a bit of rejiggering on the `audit` function, but should be largely reviewable as a standalone piece of code

## Related Issue

Fixes #961

## Acceptance Criteria

- [ ] When a eCR is saved a `ecr` `create` audit log is also emitted (when db is available)


containers/ecr-viewer/src/app/services/auditLogService.ts
```diff
@@ -32,7 +32,7 @@ type AuditableFn<
  * parameters passed to the function, and other request metadata.
  * @param subject Subject of the action being audited (e.g. "user")
  * @param action Action being done (e.g. "create")
- * @param fn Function to audit upon successful completion. Must be called with only one argument, which is an object with all of the parameters. It can return either the UUID of the subject or void. The wrapper will inject the second argument of a Kysely transaction, which should be used as the database in any queries the function executes
+ * @param fn Function to audit upon succesful completion. Must be called with only one argument, which is an object with all of the parameters. If it returns a string, it is assumed to be the `uuid` of the subject and will be added to the params as such. The wrapper will inject the second argument of a Kysely transaction, which should be used as the database in any queries the function executes
```
JNygaard-Skylight, line 35:
```suggestion
 * @param fn Function to audit upon successful completion. Must be called with only one argument, which is an object with all of the parameters. If it returns a string, it is assumed to be the `uuid` of the subject and will be added to the params as such. The wrapper will inject the second argument of a Kysely transaction, which should be used as the database in any queries the function executes
```

JNygaard-Skylight (changes requested):
Looks good, just a spelling regression.

JNygaard-Skylight (approved):
🪵

PR #1022, opened 2025-07-24
feat: Add Audit Logging to eCR Search

### LINKED ISSUE #963: Audit logging: list eCRs
## What needs to be done

Add audit logging support for eCR listing in the library

## Why it needs to be done

We need to have track of all patient record queries in the audit log

## Acceptance Criteria

- [ ] When the library page is loaded `ecr` `query` audit log is also emitted (when db is available)

## Technical Notes

JNygaard-Skylight:
# PULL REQUEST

## Summary

Add audit logging to eCR search.

## Related Issue

Fixes #963

## Acceptance Criteria

- [ ] When the library page is loaded `ecr` `query` audit log is also emitted (when db is available)

### DIFF
--- a/containers/ecr-viewer/src/app/services/listEcrDataService.ts
+++ b/containers/ecr-viewer/src/app/services/listEcrDataService.ts
@@ -107,7 +107,7 @@ export const listEcrData = audit(
 );
 
 const createSearchQuery = async (
-  db: Kysely<Core>,
+  db: Transaction<Core>,
   user: User,
   startIndex: number,
   itemsPerPage: number,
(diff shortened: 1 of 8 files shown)


containers/ecr-viewer/src/app/services/listEcrDataService.ts
```diff
@@ -57,71 +60,106 @@ export interface EcrDisplay {
...
+      itemsPerPage: number;
+      sortColumn: string;
+      sortDirection: string;
+      filterDates: DateRangePeriod;
+      searchTerm?: string;
+      filterConditions?: string[];
+    },
+    trx: Transaction<Core>,
+  ): Promise<EcrDisplay[]> => {
+    const user = await getLoggedInUser();
```
mcmcgrath13, line 86: this can now take the transaction, otherwise mssql sometimes gets sad

```suggestion
    const user = await getLoggedInUser(trx);
```

mcmcgrath13: you could also move it inot the query function since it isn't used out here and avoids adding another variable to pass

mcmcgrath13 (approved):
✅

PR #1029, opened 2025-07-25
feat: Audit Log View eCR

### LINKED ISSUE #963: Audit logging: list eCRs

JNygaard-Skylight:
# PULL REQUEST

## Summary

Adds audit logging to viewing eCRs.

## Related Issue

Fixes #963 

## Acceptance Criteria

- [ ] When a eCR is viewed a `ecr` `view` audit log is also emitted (when db is available)


containers/ecr-viewer/src/app/services/auditLogService.ts
```diff
@@ -23,7 +24,7 @@ type Action =
 
 type AuditableFn<
   Params extends Record<string, unknown>,
-  Ret extends string | boolean | void,
+  Ret extends string | boolean | FhirDataResponse | void,
```
mcmcgrath13, line 27: if we make this `any` or `unknown` do we lose anything?

mcmcgrath13:
I think we need a guard in either the audit function of in the applying of `audit` to `getFhirData` to check if wer'e using a DB. The integrated tests are failing/timing out

containers/ecr-viewer/src/app/services/auditLogService.ts
```diff
@@ -21,51 +22,86 @@ type Action =
...
+ */
+export const auditWithoutTrx = <Params extends Record<string, unknown>, Ret>(
+  subject: Subject,
+  action: Action,
+  fn: (params: Params) => Promise<Ret>,
+) => {
+  return async (params: Params): Promise<Ret> => {
+    if (!dbDialect()) {
+      return await fn(params);
+    }
```
mcmcgrath13, line 76: Could we have added this at the beginning of the original `audit` function and gotten to the same place with a lot less change? I'm waffling on whether the added functions/explicitness is worth the clarity here

containers/ecr-viewer/src/app/services/fhirDataService.ts
```diff
@@ -42,17 +44,21 @@ export const isSuccessResponse = (resp: Response): resp is SuccessResponse =>
...
-  } else if (process.env.SOURCE === GCP_SOURCE) {
-    return await getGcp(ecr_id);
-  } else {
-    return { payload: { message: "Invalid source" }, status: 500 };
-  }
-}
+export const getFhirData = audit(
+  "ecr",
+  "view",
+  async ({ ecr_id }: { ecr_id: string | null }) => {
```
mcmcgrath13, line 50:
```suggestion
  async ({ eicr_id }: { eicr_id: string | null }) => {
```

Could we switch this to `eicr_id` so it matches the other audit function param names?

containers/ecr-viewer/tests/e2e/dual/ecr-viewer.spec.ts
```diff
@@ -114,4 +117,36 @@ test.describe("viewer page", () => {
       expect(navIndex).toBe(numLinks);
     });
   });
+
+  test("audit logging", async ({ page }) => {
+    test.skip(
+      ["AWS_INTEGRATED", "AZURE_INTEGRATED", "GCP_INTEGRATED"].includes(
```
mcmcgrath13, line 123: switch this to a still-clunky endswith logic 🙏

mcmcgrath13 (approved):
🚢

PR #1030, opened 2025-07-25
refactor: Move view-data services

JNygaard-Skylight:
# PULL REQUEST

## Summary

- There are several services that are exclusively used by components within `view-data`. Therefore they can be moved into the same directory.
- removed a circular import by moving a type out of a service.

## Related Issue

Fixes # n/a


mcmcgrath13 (changes requested):
Overall a nice cleanup! I feel pretty strongly about not splitting the formatters, but otherwise looks good

containers/ecr-viewer/src/app/components/EcrFilters.tsx
```diff
@@ -35,6 +37,55 @@ interface FilterProps {
...
+ *  A custom date input component for selecting a date.
+ * @param props - The properties for the CustomDateInput component.
+ * @param props.label - The label of the custom date input component.
+ * @param props.onDateChange - The function that is called when the date changes.
+ * @param props.defaultValue - The default value of the date input.
+ * @param props.isRequired - Boolean indicating whether or not the date is required.
+ * @param props.minValue - The minimum value of the date input.
+ * @returns A JSX element containing a date input field and corresponding label.
+ */
+export const CustomDateInput = ({
```
mcmcgrath13, line 50: Could you move this somewhere below its call site in the `DateFilters`?

JNygaard-Skylight: Moved this into `FilterByDate`.

containers/ecr-viewer/src/app/view-data/services/fhirDataService.ts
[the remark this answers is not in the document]
JNygaard-Skylight: I don't think there is anything wrong with real service being in in `view-data`, as long it is only being used by the components in `view-data`.

containers/ecr-viewer/src/app/components/EcrFilters.tsx
JNygaard-Skylight: Actually, I reverted these changes because it broke tests, and I don't want to spend anymore time on this to figure it out.

containers/ecr-viewer/src/app/components/EcrFilters.tsx
```diff
@@ -216,6 +167,45 @@ const FilterReportableConditions = ({
  * - Updates the browser's query string when the filter is applied.
  */
 const FilterByDate = ({ initCustomDate, initDateRange }: FilterProps) => {
+  const CustomDateInput = ({
```
mcmcgrath13: the thinking on why this was currently in the `BaseFilter` file as it's designed to be a re-usable component. Putting it inside the `FilterByDate` really loses this. I'm okay with co-locating it file-wise given no one else is yet using it, but would really rather it not be encapsulated in another function

mcmcgrath13: looks like this was already reverted

mcmcgrath13 (approved):
🧹

PR #1034, opened 2025-07-28
fix: dupe program name error handling

### LINKED ISSUE #905: Program area form: Creating a duplicate program name does not have specific error message
## Description

Creating a duplicate program name does not have specific error message

## Steps to reproduce

1. "Create a program
2. Create a 2nd program with the same name
3. Look at error message (does not tell you why saving program area failed)."

## Expected Result

Duplicate name in program area form

## Actual result

_Explain what's happening now when the bug is present. Attach screenshot if applicable_

## Priority

#### Priority

## Notes

mcmcgrath13:
# PULL REQUEST

## Summary

Add field-level error handling for duplicate program names. Names are considered duplicate if they lower-case to the same value (now also enforced on program create/update on the db service).

## Related Issue

Fixes #905

## Additional Information

image


containers/ecr-viewer/src/app/services/programAreaService.ts
```diff
@@ -52,12 +56,34 @@ export const createProgramArea = async (
...
 
+const checkDupeName = async (
+  db: Transaction<Core>,
+  name: string,
+  uuid: string,
+) => {
+  const dupe = await db
+    .selectFrom("program_area")
+    .selectAll()
+    .where((eb) =>
```
JNygaard-Skylight, line 76: What is `eb`? Perhaps it should have a slightly more descriptive name?

containers/ecr-viewer/src/app/admin/program/ProgramForm.tsx
```diff
@@ -86,12 +88,18 @@ export const ProgramForm = ({
...
+    .filter((n) => n !== initValues.name?.toLowerCase())
+    .includes(name.toLowerCase());
+
   const initSelectedConditions = sortedCodes(initValues.conditions);
   const touched =
     (name && name !== initValues.name) ||
     selectedConditions.length !== initSelectedConditions.length ||
     selectedConditions.some((c, i) => initSelectedConditions[i] !== c);
-  const valid = !!name.trim() && numConditionsSelected > 0;
+  const valid = !!name.trim() && numConditionsSelected > 0 && !nameIsDupe;
```
akasper, line 102:
```
if (name.trim().length < 2 || conditions.length === 0) {
      throw new UserFacingError(
        "Invalid program area. Must have a non-empty name and at least one condition assigned.",
      );
    }
```
   
Since we require name length to be at least 2, can we add that to the validity check? Currently, the user can attempt to save a name of length 1 and and the error message just says it must be non-empty.

Screenshot 2025-08-01 at 12 45 59 PM

akasper (commented):
Looks good to me

PR #1035, opened 2025-07-28
feat: add auditing for create, update, delete program area

### LINKED ISSUE #960: Audit logging: Create/Edit/Delete program area
## What needs to be done

Add audit logging support for program area CRUD

## Why it needs to be done

We need to have track of all user status/roles changes in the log, which are influenced by what program areas they have access to and what those program area has access to

## Acceptance Criteria

- [ ] When a program area is deleted a `program` `delete` audit log is also emitted
- [ ] When a program area is created a `program` `create` audit log is also emitted
- [ ] When a program area is updated a `program` `update` audit log is also emitted

## Technical Notes

gordonfarrell:
# PULL REQUEST
## Summary

Applies auditing to program area creation, editing, and deletion

## Related Issue

Fixes #960

## Acceptance Criteria

- [ ] When a program area is deleted a `program` `delete` audit log is also emitted
- [ ] When a program area is created a `program` `create` audit log is also emitted
- [ ] When a program area is updated a `program` `update` audit log is also emitted


containers/ecr-viewer/tests/integration/programAreaService.test.ts
```diff
@@ -105,17 +120,30 @@ describe("program area service", () => {
...
-    await expect(createProgramArea(progName, [])).rejects.toThrow();
+    await expect(
+      createProgramArea({ name: progName, conditions: [] }),
+    ).rejects.toThrow();
   });
 
   it("should update a program area name", async () => {
     const progName = "Sad Times";
-    const id = await createProgramArea(progName, ["123"]);
+    const id = await createProgramArea({ name: progName, conditions: ["123"] });
```
mcmcgrath13, line 130:
```suggestion
    const uuiid = await createProgramArea({ name: progName, conditions: ["123"] });
```

super minor takeitorleaveit nit - if you name this uuid, then you can just do `uuid` instead of `uuid: id` in the places

mcmcgrath13 (approved):
🚀

PR #1038, opened 2025-07-29
bug: Do not show relevant lab results and clinical data section in Condition summary if no data

### LINKED ISSUE #1012: eCR Summary - When “relevant labs and clinical info” is empty, do not show those fields.
## Description

When there is no clinical information OR labs relevant to the eCR, remove those sections from the summary. 

When we were in TN for the onsite, the eCR coordinator specified that epis may write off eCRs if clinical information or lab information specifies that there is no information.

## Steps to reproduce

1. Open an eCR 
2. Look at the individual 

## Priority

#### Priority

## Notes

JNygaard-Skylight:
# PULL REQUEST

## Summary

If a condition does not have clinical data, lab reports, or immunizations do not include it in the summary.

## Related Issue

Fixes #1012 

## Old
image

## New
image

### DIFF
--- a/containers/ecr-viewer/src/app/view-data/components/EcrSummary.tsx
+++ b/containers/ecr-viewer/src/app/view-data/components/EcrSummary.tsx
@@ -53,31 +53,30 @@ const EcrSummary: React.FC<EcrSummaryProps> = ({
             {condition.conditionDetails.map((item, i) => (
               <DataDisplay item={item} key={`condition-${i}`} />
             ))}
-            {hasImmunizationDetails ||
-              (hasClinicalDetails && (
-                <>
-                  <h5
-                    className="text-bold margin-top-0 margin-bottom-1"
-                    id="relevant-clinical"
-                  >
-                    Clinical Sections Relevant to Reportable Condition
-                  </h5>
-                  {hasImmunizationDetails && (
-                    <div className="margin-top-0">
-                      {condition.immunizationDetails.map((item, i) => (
-                        <DataTableDisplay item={item} key={`imx-${i}`} />
-                      ))}
-                    </div>
-                  )}
-                  {hasClinicalDetails && (
-                    <div className="margin-top-0">
-                      {condition.clinicalDetails.map((item, i) => (
-                        <DataTableDisplay item={item} key={`detail-${i}`} />
-                      ))}
-                    </div>
-                  )}
-                </>
-              ))}
+            {(hasImmunizationDetails || hasClinicalDetails) && (
+              <>
+                <h5
+                  className="text-bold margin-top-0 margin-bottom-1"
+                  id="relevant-clinical"
+                >
+                  Clinical Sections Relevant to Reportable Condition
+                </h5>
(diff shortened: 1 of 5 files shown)


mcmcgrath13:
I read the ticket differently. That the conditions should always be there, but if there are no labs or problems or whatever, then remove that line item from the drop down. @ashton-skylight could you clarify?

JNygaard-Skylight:
> I read the ticket differently. That the conditions should always be there, but if there are no labs or problems or whatever, then remove that line item from the drop down. @ashton-skylight could you clarify?

Sorry, reread the ticket, I see what we want to do. When a section is empty, we want to just not display it instead of displaying "No matching...".

containers/ecr-viewer/src/app/view-data/components/EcrSummary.tsx
```diff
@@ -51,35 +53,49 @@ const EcrSummary: React.FC<EcrSummaryProps> = ({
...
-              className="text-bold margin-0 margin-bottom-1"
-              id="relevant-labs"
-            >
-              Lab Results Relevant to Reportable Condition
-            </h5>
-            <div className="margin-top-0">
-              {condition.labDetails.map((item, index) => (
-                <DataTableDisplay item={item} key={`${item.title}-${index}`} />
+            {hasImmunizationDetails ||
+              (hasClinicalDetails && (
```
angelathe, line 57: Why place the `(` before `hasClinicalDetails`?

JNygaard-Skylight:
> @mcmcgrath13 @JNygaard-Skylight Might've interpreted this differently, but I thought @ashton-skylight said that the whole section should be removed? Out of worry that folks might not actually open the accordion to see that there are no relevant clinical data/labs for that condition
> 
> But your work matches the expected result so 🤷‍♀️

This was also my assumption, but see @mcmcgrath13's comment. I'm now pretty unsure of myself, so I'll wait until Ashton has a look.

mcmcgrath13:
> > @mcmcgrath13 @JNygaard-Skylight Might've interpreted this differently, but I thought @ashton-skylight said that the whole section should be removed? Out of worry that folks might not actually open the accordion to see that there are no relevant clinical data/labs for that condition
> > But your work matches the expected result so 🤷‍♀️
> 
> This was also my assumption, but see @mcmcgrath13's comment. I'm now pretty unsure of myself, so I'll wait until Ashton has a look.

I pinged @ashton-skylight to weigh in here 😅 

I thought the concern was that if people opened the accordion and so no relevant clinical info was found, they'd take that as base truth and not go look around the eCR for more details. So we didn't want to assert there's no info, but still want to highlight that this was a condition on the eCR

containers/ecr-viewer/src/app/view-data/components/EcrSummary.tsx
JNygaard-Skylight: I do not need those brackets, but I do added brackets around the or statement. That will also fix the snapshot.

angelathe (approved):
🎉🎉

PR #1039, opened 2025-07-30
refactor: Harden FHIR Paths

### LINKED ISSUE #932: Remove `.with(meta.profile="...` from FHIR paths
## What needs to be done

Replace FHIR paths with `.with(meta.profile="...` with something else.

## Why it needs to be done

For many of our FHIR paths we use the `meta.profile` of a resource to find a resource of a particular type. As an example:
```yaml
patientPregnancyStatus: {
  type: "ValueX",
  path: "Bundle.entry.resource.where(resourceType = 'Observation').where(meta.profile = 'http://hl7.org/fhir/us/ecr/StructureDefinition/us-ph-pregnancy-status-observation').value",
},
```
However, while when we convert eICRs we, most of the time, do add `meta.profile`, it is not a required field. If the eCR Viewer is expected to ingest arbitrary FHIR, it can not rely on `meta.profile`. 

## Acceptance Criteria

- All FHIR paths used by the eCR Viewer use elements of resources that we can guarantee exist.

## Technical Notes

Most resources (I think)  should have some other way to identify them. As an example: The [Pregnancy Status Observation](https://hl7.org/fhir/us/ecr/2.1.2/StructureDefinition-us-ph-pregnancy-status-observation.html) resource must have the following code:
```json
"code" : {
  "coding" : [
    {
      "system" : "http://loinc.org",
      "code" : "82810-3",
      "display" : "Pregnancy status"
    }
  ]
},
```
However for other resources it might be more complicated. As an example a [RR eICR Processing Status Observation](https://hl7.org/fhir/us/ecr/2.1.2/StructureDefinition-rr-eicr-processing-status-observation.html) can have a `code` of any value within the [eICR Processing Status valueset](https://phinvads.cdc.gov/vads/ViewValueSet.action?oid=2.16.840.1.113883.10.20.15.2.5.8).

## Additional context

It's possible that we can in fact assume `meta.profile` exists, I'm just going over my interpretation of the IG.

JNygaard-Skylight:
# PULL REQUEST

## Summary

Changed many FHIR paths to search for a particular `coding` instead of either look for the `meta.profile`, which is typically not required, or just looking for the `code`, which technically could mean anything without a system.

Several FHIR paths that we use were the `meta.profile` to find the appropriate resource. While this is correct, it is typical not a required field of the resource.

## Related Issue

Fixes #932

### DIFF
--- a/containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
+++ b/containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
@@ -257,7 +257,7 @@ const _fhirPathMappings: { [K in FhirPathKeys]: Omit<FhirPath<K>, "name"> } = {
   },
   patientOccupationHistory: {
     type: "Observation",
-    path: "entry.resource.Observation.where(meta.profile = 'http://hl7.org/fhir/us/odh/StructureDefinition/odh-PastOrPresentJob')",
+    path: "entry.resource.Observation.where(code.coding.exists(system = 'http://loinc.org' and code = '11341-5'))",
   },
   patientEmploymentStatus: {
     type: "Observation",
--- a/containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
+++ b/containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
@@ -265,19 +265,19 @@ const _fhirPathMappings: { [K in FhirPathKeys]: Omit<FhirPath<K>, "name"> } = {
   },
   patientTobaccoUse: {
     type: "ValueX",
-    path: "entry.resource.Observation.where(code.coding.code = '72166-2').where(category.coding.code = 'social-history').value",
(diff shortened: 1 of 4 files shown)


containers/ecr-viewer/src/app/utils/evaluate/fhir-paths.ts
```diff
@@ -435,11 +435,11 @@ const _fhirPathMappings: { [K in FhirPathKeys]: Omit<FhirPath<K>, "name"> } = {
   // Clinical Info
   clinicalReasonForVisit: {
     type: "ValueX",
-    path: "entry.resource.section.where(title.lower() = 'reason for visit').extension.value",
+    path: "entry.resource.Composition.section.where(code.coding.exists(system = 'http://loinc.org' and code = '29299-5')).extension.value",
```
mcmcgrath13, line 438: is it weird here that we're just looking for any extension's value?

JNygaard-Skylight: Fixed.

PR #1060, opened 2025-08-06
fix: don't enable filter apply buttons unless state has changed

### LINKED ISSUE #57: Modify interaction states and design for Apply Filter buttons
## What needs to be done

Right now, the `Apply Filter` buttons for `Filter by Date` and `Filter by Condition` are always enabled. We need to make sure that the button is only enabled when a filter change has been made. We also want to update the designs of the button so that it matches the designs for the disabled/enabled/hover/active states of the button.

## Why it needs to be done

1. `Apply Filter` button should only be enabled when a filter change has been made
    - **Why?** User need clear feedback when they have made a pending modification to the filter they are interacting with.
2. Update the `Apply Filter` button designs for disabled/enabled/hover/active state of the button
    - **Why?** There should be distinct visual cues for each state to improve usability -- it makes the button's status clear to users.

## Design review
Apply filter buttons 
[Link to Figma in dev mode](https://www.figma.com/design/zAPyOEYwNtx3f2Y47aoq0G/eCR-Viewer-(Aug-Dec-2024)?node-id=8349-24999&m=dev)

## Acceptance Criteria
- [ ] For each filter box (by date and by condition), the `Apply Filter` button should only be enabled when a change has been made
  1. A condition has been checked/unchecked
  2. A different pre-defined date range has been selected
  3. A different valid custom date range has been entered
  - [ ] The button should be enabled whether or not it ends up getting clicked. If the page is re-loaded and the query is re-set to its previous state, the `Apply Filter` button should be reset to the `disabled` state
- [ ] The Apply Filter button must have distinct visual styles for the following states:
    - [ ] Disabled: Button is not clickable
    - [ ] Enabled: Button is clickable
    - [ ] Hover: Button is hovered over
    - [ ] Active: Button is clicked

## Technical Notes

**Relevant files:**
- `containers/ecr-viewer/src/app/components/Filters.tsx`

## Additional context
N/A

mcmcgrath13:
# PULL REQUEST

## Summary

Only enable the "Apply filter" buttons once the filter's state has changed

* Add `touched` prop to the `Filter` component (optional, since not all filters use the apply buttons
* Use the `touched` prop on the ecr library filters

## Related Issue

Fixes #57

## Acceptance Criteria

- [ ] For each filter box (by date and by condition), the `Apply Filter` button should only be enabled when a change has been made
  1. A condition has been checked/unchecked
  2. A different pre-defined date range has been selected
  3. A different valid custom date range has been entered
  - [ ] The button should be enabled whether or not it ends up getting clicked. If the page is re-loaded and the query is re-set to its previous state, the `Apply Filter` button should be reset to the `disabled` state
- [ ] The Apply Filter button must have distinct visual styles for the following states:
    - [ ] Disabled: Button is not clickable
    - [ ] Enabled: Button is clickable
    - [ ] Hover: Button is hovered over
    - [ ] Active: Button is clicked

## Additional Information

image

image


mcmcgrath13:
> > > Looks good. One thing I'm seeing is that when the end date of the custom date range is left blank, the apply button is still enabled. When clicked, it closes the filter, but has no effect on the table. Reopening the filter and setting the end date to today disables the button.
> > 
> > 
> > I've fixed the issue where it enables when the option is custom and the start/end haven't changed. When I follow the flow you outline, when I save with no end date, today is defaulted in to the end date, so on re-opening the end is today's date, so it is disabled if still set to today. Does that match what you see?
> 
> Here are the steps:
> 
> 1. Open date date filter and apply a non-custom date range (e.g. Last year).
> 2. Reopen the filter. Select Custom date range without an end date. Apply.
> 3. Open filter again and see that today is now the end date.
> 4. Delete the end date. Apply is now enabled. Click apply. The filter will close and the table is not updated.
> 5. Open filter. The end date is still empty and the apply button is enabled (and has no effect on the table when clicked).
> 6. Set the end date to today. The apply button becomes disabled.

Nice find!! This was a bug with the state logic, so fixed that up and this seems to be consistent now

akasper (approved):
:shipit:

PR #1064, opened 2025-08-06
feat: add GCP as an option in the deployment environment and remove t…

alismx:
## Changes Proposed

- Added gcp trigger support

## Reviews

- Verify that the new GCP option appears correctly in the workflow's available choices.
- Confirm that removing `trigger-vm-workflow.yaml` does not break existing workflows or integrations.


mcmcgrath13:
I see that the job triggered as I'd expect from the ecr-viewer side, but went to look at the infra side and don't see it at https://github.com/skylight-hq/dibbs-tf-envs/actions as the job says it should be - is that expected?

alismx:
> I see that the job triggered as I'd expect from the ecr-viewer side, but went to look at the infra side and don't see it at [skylight-hq/dibbs-tf-envs/actions](https://github.com/skylight-hq/dibbs-tf-envs/actions) as the job says it should be - is that expected?

My PR hasn't been merged on that side, yet. It should be soon.

mcmcgrath13 (approved):
thank you!

PR #1089, opened 2025-08-14
fix: reset condition search term when library filter resets

mcmcgrath13:
# PULL REQUEST

## Summary

What changes are being proposed?

Fix minor UX bug I noticed where the search term wasn't reseting with the condition filter closed


akasper (approved):
Looks good :shipit:

PR #1097, opened 2025-08-18
build: only specify python image versions to the minor

mcmcgrath13:
# PULL REQUEST

## Summary

Remove patch specifiers on python images

(https://skylight-hq.slack.com/archives/C05B5R8K0TF/p1755521933757759)


mcmcgrath13:
We've decided that specifying to the patch is our preference, so closing this out

PR #1124, opened 2025-09-05
fix: use encounter in composition

### LINKED ISSUE #1122: Use encompassing encounter

JNygaard-Skylight:
# PULL REQUEST

## Summary

Now that the FHIR converter is converting planned encounters into their own `Encounter` resources. It is now possible to have multiple `Encounter` resources in a bundle. Therefore we need to explicitly get the `encounter` referenced in the `Composition` where we used to just get the first `Encounter`.

## Related Issue

Fixes #1122


JNygaard-Skylight:
Waiting on changes in #1116 to make changes to the message-parser

PR #1141, opened 2025-09-12
major: set default schema for kysely migration tables

### LINKED ISSUE #1102: Move Kysely Migration tables to ecr_viewer schema
## What needs to be done

Kysely creates two tables automatically to track migrations. Without customization these are created in the default schema of the DB, e.g. DBO. We need to change this behavior so these tables are colocated with the rest of our viewer tables and we don't have random (critical) tables floating around elsewhere

We need to coordinate with devops to move ecr-viewer schema creation out of our migrations and into infrastructure setup so its available for kysely to work with from the start.

This will also require a manual script for existing users to run that copies data from their two migration tables in the default schema into a new table in the ecr_viewer schema.

We also need to customize our kysely migrator to specify that these tables should be created under the `ecr_viewer` schema rather than the default

And write clear instructions for how to implement these changes if its not a fresh install

## Why it needs to be done

We ran into issues with this on a user call where a STLT had deleted all tables in DBO assuming everything for the viewer was under ecr_viewer, which caused the app to think no migrations had yet been run when they had. The only solution was to manually repair with copied data from another DB since trying to run migrations again could cause issues

## Acceptance Criteria
- [ ] Create a script (or scripts, this needs to work for both DB types) for upgrading users to run manually which copies the data from both of their default schema migration tables into a new version under the `ecr_viewer` schema.
- [ ] Coordinate with devOps to create the `ecr_viewer` schema during the infrastructure setup stage
- [ ]  Update migrations to no longer create schema, and change the initial migration to throw if the `ecr_viewer` schema does NOT exist
- [ ] Update the getMigrator function in migrate.ts to specify that `migrationTableName` and `migrationLockTableName` fall under the `ecr_viewer` schema
- [ ] Write clear documentation for STLTs making this upgrade to follow

## Technical Notes
This is a BREAKING change. New installs should have zero problems, but ALL upgrading STLT installs using a DB will need to make sure the new version of the table under the ecr_viewer schema is created and populated properly or they could run into serious issues.

austin-hall-skylight:
# PULL REQUEST

## Summary

Sets default schema for Kysley migration tables to match the schema of the other tables we are creating instead of writing them to the database default. Removes schema creation from init migration.

## Related Issue

Fixes #1102

## Acceptance Criteria

- [x] Create a script (or scripts, this needs to work for both DB types) for upgrading users to run manually which copies the data from both of their default schema migration tables into a new version under the `ecr_viewer` schema.
- [ ] Coordinate with devOps to create the `ecr_viewer` schema during the infrastructure setup stage
- [x]  Update migrations to no longer create schema, ~and change the initial migration to throw if the `ecr_viewer` schema does NOT exist~ changing the migration to throw if the schema doesn't exist doesn't work because Kysely will throw an error first when it tries to create the migration tables in the `ecr_viewer` schema
- [x] Update the getMigrator function in migrate.ts to specify that `migrationTableName` and `migrationLockTableName` fall under the `ecr_viewer` schema
- [x] Write clear documentation for STLTs making this upgrade to follow

## Additional Information

## Checklist


austin-hall-skylight:
Waiting for [this PR](https://github.com/kysely-org/kysely/pull/1608) to be merged and released before this can be completed.

laura-puerto-skylight (approved):

PR #1197, opened 2025-10-17
fix: increase request timeout from eCR viewer to orchestration

### LINKED ISSUE #1196: Increase timeout for requests from viewer to orchestration
## What needs to be done
Increase the request timeout for requests from the viewer to the orchestration service originating from `/process-ecr`.

## Why it needs to be done
When the FHIR converter takes a long time to respond, either because the eCR being converted is large or because of heavy traffic, the request from the eCR Viewer to orchestration times out. This causes errors that could be avoided if the eCR was given more time to go through the pipeline.

## Acceptance Criteria

GIVEN `/process-ecr` is called
IF the eCR takes more than 5 minutes to process
THEN the request will not time out

austin-hall-skylight:
# PULL REQUEST

## Summary

Increase the request timeout for requests from the viewer to the orchestration service originating from `/process-ecr`.

## Related Issue

Fixes #1196 

## Acceptance Criteria

GIVEN `/process-ecr` is called
IF the eCR takes more than 5 minutes to process
THEN the request will not time out

## Additional Information

For testing, I added `Thread.Sleep(300000);` to the beginning of the `/convert-to-fhir` handler function in the FHIR converter.

## Checklist

### DIFF
--- a/.github/workflows/create-new-release.yaml
+++ b/.github/workflows/create-new-release.yaml
@@ -145,7 +145,7 @@ jobs:
           ref: docs
 
       - name: Download container docs from artifacts
-        uses: actions/download-artifact@v5.0.0
+        uses: actions/download-artifact@v6.0.0
         with:
           pattern: container-docs-*
           merge-multiple: true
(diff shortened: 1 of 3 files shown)


gordonfarrell (approved):
LGTM! I like this new pattern using the mockPool, its really clear and readable

PR #1213, opened 2025-11-05
feat: saving of original xml files

### LINKED ISSUE #1099: Create configurable option to save raw eICR/RR data
## What needs to be done

Create a way for users to save the raw, original eICR/RR data in our data containers along with the FHIR bundles.

## Why it needs to be done

Philadelphia has requested this feature - they want to be able to open the original eICR/RR along with the Viewer, so they can double-check that everything is present.

## Acceptance Criteria

- [ ] eICR and RR XMLs can be saved to blob storage across all the cloud providers 
- [ ] Environment variable exists to control this functionality (by default, this shouldn't be enabled)
   - [ ] Environment variables have also been added to all the required TF repos:
       - [ ] AWS TF module
       - [ ] Azure TF
       - [ ] VM (blank defaults will be needed)
- [ ] Ensure xml files are removed from blob storage if ecr fails to be processed

Work with the DevOps team to ensure you've added the environment variable to all needed locations - it may need to be set in some unexepected ones, like the VM wizard script.
       

## Technical Notes
After some preliminary discussion, we've decided **not** to have this data stored in a separate bucket from the FHIR bundles - they'll all use the same storage. We do need to decide on a naming/folder schema within the blob storage container that will allow us to store all these files; it should be backwards-compatible with our current storage methods.

## Additional context
This is only the backend and data storage work. The actual display of this data will be done in #1091.

gordonfarrell:
# PULL REQUEST
## Summary
Adds an environment variable flag which, when true, saves a zipped copy of the original XML upload to s3/Blob storage for later retrieval. It will also do a rollback delete of the saved zips if the rest of eCR processing fails.

## Related Issue

Fixes #1099

## Acceptance Criteria

- [ ] eICR and RR XMLs can be saved to blob storage across all the cloud providers 
- [ ] Environment variable exists to control this functionality (by default, this shouldn't be enabled)
   - [ ] Environment variables have also been added to all the required TF repos:
       - [ ] AWS TF module
       - [ ] Azure TF
       - [ ] VM (blank defaults will be needed)
- [ ] Ensure xml files are removed from blob storage if ecr fails to be processed

## Additional Information

Let me know if you need a hand/advice testing this out, the behavior of our various storage emulators can be kind of strange

### DIFF
--- a/containers/ecr-viewer/package.json
+++ b/containers/ecr-viewer/package.json
@@ -57,7 +57,7 @@
     "motion": "^12.23.22",
     "mssql": "^11.0.1",
     "next": "^15.4.5",
-    "next-auth": "^4.24.11",
+    "next-auth": "^4.24.12",
     "node-html-parser": "^7.0.1",
     "pg": "^8.15.6",
     "postcss-discard-comments": "^7.0.4",
--- a/containers/ecr-viewer/package-lock.json
+++ b/containers/ecr-viewer/package-lock.json
@@ -28,7 +28,7 @@
         "motion": "^12.23.22",
         "mssql": "^11.0.1",
         "next": "^15.4.5",
-        "next-auth": "^4.24.11",
+        "next-auth": "^4.24.12",
         "node-html-parser": "^7.0.1",
         "pg": "^8.15.6",
         "postcss-discard-comments": "^7.0.4",
@@ -15082,9 +15082,9 @@
       }
     },
     "node_modules/next-auth": {
-      "version": "4.24.11",
-      "resolved": "https://registry.npmjs.org/next-auth/-/next-auth-4.24.11.tgz",
-      "integrity": "sha512-pCFXzIDQX7xmHFs4KVH4luCjaCbuPRtZ9oBUjUhOk84mZ9WVPf94n87TxYI4rSRf9HmfHEF8Yep3JrYDVOo3Cw==",
+      "version": "4.24.12",
+      "resolved": "https://registry.npmjs.org/next-auth/-/next-auth-4.24.12.tgz",
+      "integrity": "sha512-wooJAL5Md9Fn2UwUI2qN9TY/+k8HJGRyi3TdSt/xHfDTtdpPxDqmo4v8hUrKGb+d66FB/rYy9RutA/9EeJrK0Q==",
       "license": "ISC",
       "dependencies": {
         "@babel/runtime": "^7.20.13",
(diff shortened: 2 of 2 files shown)


containers/ecr-viewer/src/app/api/save-fhir-data/service.ts
```diff
@@ -59,16 +70,25 @@ export const saveFhirData = async (
...
 ): Promise<SaveResponse> => {
-  const objectKey = `${ecrId}.json`;
+  let objectKey = "";
+
+  if (fileType === "fhir") {
+    objectKey = `${ecrId}.json`;
+  } else if (fileType === "xml") {
+    objectKey = `${ecrId}.zip`;
+  }
+  console.log("About to delete");
```
austin-hall-skylight: I think this can be deleted.

containers/ecr-viewer/src/app/data/blobStorage/s3Client.ts
```diff
@@ -47,23 +47,30 @@ export const s3HealthCheck = async () => {
...
+
+    const contentType = isBuffer ? "application/zip" : "application/json";
+
     const input = {
       Body: body,
       Bucket: bucketName,
       Key: objectKey,
-      ContentType: "application/json",
+      ContentType: contentType,
+      ContentLength: Buffer.isBuffer(body)
```
austin-hall-skylight:
```suggestion
      ContentLength: isBuffer
```

containers/ecr-viewer/src/app/api/process-ecr/route.ts
```diff
@@ -69,12 +81,28 @@ export const POST = async (
 
   try {
     const { return_fhir_bundle, ...body } = routeSchema.parse(rawBody);
+
+    if (process.env.SAVE_XML) {
+      ecrId = await getEcrIdFromXml(body);
+      await zipAndSaveXml(body, ecrId);
```
austin-hall-skylight: I would suggest wrapping these and the orchestration request in Promise.All so you aren't waiting on zipping and saving before making the request to orchestration. I have some concerns about XML to JSON conversion and zipping/unzipping times for larger XMLs and I think this is one way we alleviate some of the potential performance impact.

gordonfarrell: Good idea! I ended up going with promise.allSettled to be able to more explicitly tell what failed how and trigger the rollback properly if needed

containers/ecr-viewer/package.json
```diff
@@ -41,14 +41,17 @@
     "@fullhuman/postcss-purgecss": "^6.0.0",
     "@google-cloud/storage": "^7.17.2",
     "@next/bundle-analyzer": "^15.3.1",
+    "@parcel/watcher": "^2.5.0",
```
austin-hall-skylight: This looks like it's not being used anywhere, can we delete it?

PR #1430, opened 2026-04-27
feat: Adding optional HA Proxy for FHIR Converter

laura-puerto-skylight:
## Summary

Adding optional load balancing service that uses HA Proxy to distribute load to the FHIR converters using a "least connections" algorithm to more evenly distribute load vs the built-in Round Robin algorithm used by AWS Service Connect. 

## Related Issue

Fixes #1369 

## Acceptance Criteria

- [x] Reproduce issue in test environment
- [x] Get HA Proxy working on local environment
- [x] Get HA Proxy working on test environment
- [x] Validate issue is mitigated in test env

## Additional Information

Additional infrastructure PRs:

- https://github.com/CDCgov/terraform-aws-dibbs-ecr-viewer/pull/54

PR #1435, opened 2026-04-28
feat: add eCR Viewer VM deployment scripts and README rewrite

alismx:
## Summary
Adds VM deployment scripts for the eCR Viewer stack and rewrites README.md (164 -> 78 lines).

## Changes
- **deployment/vm/*.sh** - 6 scripts: ecrv-init.sh (downloader), docker-install.sh, ecrv-update.sh + ecrv-wizard.sh (config), config-backup.sh, apt-updates.sh
- **docker-compose.yaml** - 6-service stack
- **README.md** - rewritten: fixed ecrw- -> ecrv- typos, removed bloat, condensed descriptions
- **check.sh** - shellcheck tests, all but 2 pass. Future work.
- **deployment/vm/docker-compose.yaml**: Added `depends_on` with `condition: service_healthy` to the orchestration service

## Why
Reproducible deployment path for eCR Viewer multi-service stack on Ubuntu VMs.

## TODO

- [x] Update wget urls to use `main` branch.

## Testing
- [x] Manual deploy on Ubuntu 24.04+ VM
- [x] Verify wizard flow with config profiles

### DIFF
--- a/deployment/vm/dibbs-orchestration.env
+++ b/deployment/vm/dibbs-orchestration.env
@@ -1,0 +1,7 @@
+OTEL_METRICS=none
+OTEL_METRICS_EXPORTER=none
+ECR_VIEWER_URL=http://dibbs-ecr-viewer:3000/ecr-viewer
+INGESTION_URL=http://dibbs-ingestion:8080
+FHIR_CONVERTER_URL=http://dibbs-fhir-converter:8080
+MESSAGE_PARSER_URL=http://dibbs-message-parser:8080
+TRIGGER_CODE_REFERENCE_URL=http://dibbs-trigger-code-reference:8080
--- a/deployment/vm/docker-compose.yaml
+++ b/deployment/vm/docker-compose.yaml
@@ -1,0 +1,107 @@
+services:
+  ecr-viewer:
+    image: ghcr.io/cdcgov/dibbs-ecr-viewer/ecr-viewer:$DIBBS_VERSION
+    container_name: dibbs-ecr-viewer
+    restart: always
+    env_file: "dibbs-ecr-viewer.env"
+    ports:
+      - "3000:3000"
+    networks:
+      - dibbs
+    healthcheck:
+      test:
+        ["CMD", "wget", "--spider", "http://127.0.0.1:3000/api/health-check"]
+      interval: 5s
+      timeout: 2s
+      retries: 5
+      start_period: 30s
+
+  ingestion:
+    image: ghcr.io/cdcgov/dibbs-ecr-viewer/ingestion:$DIBBS_VERSION
+    container_name: dibbs-ingestion
+    restart: always
+    ports:
+      - "8080:8080"
+    networks:
+      - dibbs
+    healthcheck:
+      test: ["CMD", "wget", "--spider", "http://127.0.0.1:8080/"]
+      interval: 5s
+      timeout: 2s
+      retries: 5
+      start_period: 10s
+
+  fhir-converter:
+    image: ghcr.io/cdcgov/dibbs-ecr-viewer/fhir-converter:$DIBBS_VERSION
+    container_name: dibbs-fhir-converter
+    restart: always
+    ports:
+      - "8082:8080"
+    networks:
+      - dibbs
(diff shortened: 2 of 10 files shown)


deployment/vm/dibbs-orchestration.env
```diff
@@ -0,0 +1,7 @@
+OTEL_METRICS=none
+OTEL_METRICS_EXPORTER=none
+ECR_VIEWER_URL=http://dibbs-ecr-viewer:3000/ecr-viewer
+INGESTION_URL=http://dibbs-ingestion:8080
+FHIR_CONVERTER_URL=http://dibbs-fhir-converter:8080
```
laura-puerto-skylight, line 5: How are you planning on handling the proxy in these scripts? do we need a TODO here to update the FHIR Converter URL?

deployment/vm/docker-compose.yaml
laura-puerto-skylight, line 1: I think some services depend on others to be up and running like the orchestration service depends on the others. The proxy will also need the FHIR converter to be up and healthy so it can start up properly. You can use the `depends_on:` field for that (see the [HAProxy PR](https://github.com/CDCgov/dibbs-ecr-viewer/pull/1430/changes#diff-d6ada02452009973f315030ba03d809288767b764c27eb24d042bcdb7bca2e36))

deployment/vm/ecrv-init.sh
```diff
@@ -0,0 +1,119 @@
...
+
+  echo "Prerequisites check complete."
+  echo ""
+}
+
+download_scripts() {
+  cd "$HOME"
+  echo "Downloading ecrv scripts..."
+
+  wget -O apt-updates.sh ${REPO_URL}/alis/deployment_vm/deployment/vm/apt-updates.sh
```
laura-puerto-skylight: Should there be a TODO here to update these to main?

deployment/vm/dibbs-orchestration.env
alismx: @laura-puerto-skylight If/when we add the proxy to this dockerfile, we can update this .env file with the new target.

This use case, and the fact that this file isn't modified by the wizard script is a good reason to get rid of the "no clobber" in the wget of this file. I'll update that.

Let's coordinate with our partners after this and your proxy PR goes in on whether we should include the proxy and URL update in this vm setup.

deployment/vm/ecrv-init.sh
alismx: Yes!

deployment/vm/docker-compose.yaml
alismx: I see your examples, to date I haven't worried so much about start order, but I'd be happy to include that in this PR. Do we have a source of truth on preferred start order documented somewhere?

laura-puerto-skylight: I am not sure there is a source of truth, all I know is from the existing docker compose files and my experience with the proxy. We can add that for the proxy when it comes online since it is required there. For the other ones it might be ok if they come up all at once... I think

austin-hall-skylight: The viewer has dependencies on blob storage and the database (if applicable) in our docker compose for seeding data locally as well as orchestration having dependencies on the fhir converter etc. This is just to make sure that all the services are running before we send data. 

https://github.com/CDCgov/dibbs-ecr-viewer/blob/2a29662c3f95c9a73c3b2ee6861f78a397680dac/containers/ecr-viewer/docker-compose.yaml#L60

https://github.com/CDCgov/dibbs-ecr-viewer/blob/2a29662c3f95c9a73c3b2ee6861f78a397680dac/containers/ecr-viewer/docker-compose-full.yaml#L39

alismx: I'll push an update adding `depends_on` settings to the orchestration service.

alismx: @austin-hall-skylight Not sure I can do anything about the storage(s3, blob) dependency.

laura-puerto-skylight (approved):
Seems ok to me!

PR #1443, opened 2026-05-04
Add new middleware to handle epitrax auth

### LINKED ISSUE #1438: Add Epitrax Authenthication method
### Description:
  When EpiTrax navigates a user to the ECR Viewer, the ECR ID is visible in the
   browser URL. This is a security concern because it exposes internal         
  identifiers that could allow users to  guess other ECR IDs.
                                                                               
 ### Goal:           
  The ECR ID should not appear in the browser address bar at any point during
  an EpiTrax-authenticated session. The auth token needs to contain the ECR ID,
   so it can be used as the source of truth instead of a URL parameter.
                                                                               
###  Acceptance Criteria:
  - [ ] The ECR ID is not visible in the browser URL at any point after navigating
  from EpiTrax                                                                 
  - [ ] The correct ECR record is still loaded for the authenticated user

vochaparro2:
# PULL REQUEST

Hide ECR ID from URL for EpiTrax-authenticated sessions 

## Summary

  - Adds withEpitraxAuth middleware that handles authentication for            
  EpiTrax-originated sessions on the /view-data page
  - On first visit, verifies the JWT passed via ?auth=, stores the auth token  
  and ECR ID in httpOnly cookies, and redirects to a clean URL with no ?id= in 
  the address bar
  - On subsequent visits, verifies the JWT and internally rewrites the request 
  so the page receives the ECR ID without it ever appearing in the browser URL 
  - Adds unit tests covering all auth paths and edge cases

## Related Issue

Fixes #1438 

## Acceptance Criteria

  - [ ] The ECR ID is not visible in the browser URL at any point after navigating
  from EpiTrax                                                                 
  - [ ] The correct ECR record is still loaded for the authenticated user

## Additional Information

## Checklist


austin-hall-skylight:
I would prefer to unify this new middleware and withNbsAuth into something like withJwtAuth. We could change the cookie names and environment variable names to be more generic. I think it would be fine to remove the eCR ID from the URL when launching from NBS like EpiTrax wants and as for verifying the eCR ID in the token, we can just skip that if there isn't an ID in the token which would be the NBS flow. We would also take on the JWT authentication for `/api/` routes from NBS which is something that we would need for purely EpiTrax-integrated instances. I'm happy to discuss this further if needed.

containers/ecr-viewer/src/app/error/auth/page.tsx
```diff
@@ -14,9 +14,10 @@ const ErrorAuthPage = () => (
   >
     Please try the following:
     <ul className="margin-0 padding-left-3">
-      {process.env.NBS_PUB_KEY && (
+      {process.env.JWT_PUB_KEY && (
```
austin-hall-skylight: Let's mark NBS_PUB_KEY and NBS_API_PUB_KEY as deprecated but keep them working for the jurisdictions that are using them. Once we've confirmed that everyone has switched over we can remove them. Apologies for not including that detail in my comment earlier.

```suggestion
      {(process.env.JWT_PUB_KEY || process.env.NBS_PUB_KEY) && (
```

austin-hall-skylight (approved):

PR #1470, opened 2026-05-14
feat: Return 409 early when duplicate ECR is detected in DB

### LINKED ISSUE #1332: Return 409 for duplicate eCR before sending it through the pipeline

vochaparro2:
# PULL REQUEST

Return 409 early when duplicate ECR is detected in DB

## Summary

Before forwarding to orchestration, check if an ECR with the same ID already exists in the metadata DB and short-circuit with 409 if so.

## Related Issue

Fixes #1332 

## Acceptance Criteria

IF I send an eCR to /process-ecr
AND the ID of that eCR already exists in the ecr_id column of the ecr_data table
THEN I will return a 409 without sending the eCR to the orchestration service
AND the load testing script will now send eCRs to dev7 instead of dev42

## Additional Information

## Checklist

### DIFF
--- a/containers/ecr-viewer/src/app/api/process-ecr/service.ts
+++ b/containers/ecr-viewer/src/app/api/process-ecr/service.ts
@@ -12,6 +12,8 @@ import {
   BundleExtendedMetadata,
   BundleMetadata,
 } from "@/app/api/save-fhir-data/types";
+import { getDb } from "@/app/data/metadataDb/database";
+import { Core } from "@/app/data/metadataDb/types/core";
 import { dbDialect, dbSchema } from "@/app/data/metadataDb/utils/db-config";
 import { getEcrIdFromIdentifier, resolveEcrId } from "@/app/utils/ecrid-utils";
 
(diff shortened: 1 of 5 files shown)


containers/ecr-viewer/src/app/api/process-ecr/service.ts
```diff
@@ -182,6 +184,22 @@ export const orchestrationRequest = async (
   returnBundle: boolean = false,
   fetchAgent = createOrchestrationAgent(),
 ) => {
+  if (dbDialect()) {
```
austin-hall-skylight, line 191: Can you refactor things so that we only have to get the eCR ID once? With the code the way it is now, we would get it once to check if the eCR has already processed and then again if the environment variable is set to save the XML. Depending on how the eCR is sent this could mean having to unzip and read the file twice.

Also I would like this check to happen before we save the XML so that we don't make the call to save if we don't need to. I think a lot of the logic in the route (approximately lines 85-117) should be moved into the service to accommodate these changes.

austin-hall-skylight (changes requested):

austin-hall-skylight (approved):

HPatelTN:
Will this work for integrated version of eCR Viewer? @austin-hall-skylight  @vochaparro2

austin-hall-skylight:
@HPatelTN this behavior will be added to the integrated version in a separate ticket. It's coming up soon though.

PR #1477, opened 2026-05-18
fix: Adding mitigations for supply chain attacks

laura-puerto-skylight:
## Summary

This adds mitigation steps for supply chain attacks in both Node and Python containers.

### Node (NPM) 

1. Adds a cooldown for packages that are more recent than 10 days, usually compromised packages are taken down during this window. 
2. Adds ignore-scripts which blocks all automatic lifecycle scripts from running during package installations, this includes adding the @lavamoat/allow-scripts package to securely allow-list packages that can run scripts. Note: New dependencies where manually cross-referenced with known compromised package lists and I also ran the checker tool to confirm.

Allowed scripts:

- parcel/watcher#2.5.6 - uses a postinstall script to figure out if you are on Mac, Windows, or Linux, and then compiles or downloads the correct native bindings
- eslint-config-next>eslint-import-resolver-typescript>unrs-resolver#1.11.1 - It uses a tool called napi-postinstall to check your environment and download the correct pre-compiled Rust binary for your system.
- sharp#0.34.5 - Package for processing images. Install script is responsible for downloading a binary so it can process images.

Not allowed scripts:
- lavamoat/preinstall-always-fail#3.0.0 - test script to make sure the package works, set to true for the build to fail
- fhirpath#4.8.5 - install script installs node modules for a demo, probably for local dev of this package but not needed for using the package itself

### Python (PIP)

1. Only use wheels (`.whl` files) which is a precompiled, zipped format of a Python package. Installing a wheel does **not** execute the `setup.py` or any arbitrary code. This is done by adding `--only-binary` to the `pip install` command. Packages that do not have a wheel are allow-listed explicitly using `--no-binary`. 
2. Adds strict hash checking by locking the hash of the wheels and enforce hash checking with `--require-hashes` in the `pip install` command.

### README Updates
Each container has instructions on how to install packages so that they work given these new mitigations. 

## Related Issue

Fixes #

## Acceptance Criteria

## Additional Information

## Checklist


angelathe (commented):
Do we think there will be a lot of manual overhead using `ignore-scripts=true` (postinstall/setup steps)? I think we should include it, but slightly worried debugging could be painful if any package setups are incomplete

laura-puerto-skylight:
> Do we think there will be a lot of manual overhead using `ignore-scripts=true` (postinstall/setup steps)? I think we should include it, but slightly worried debugging could be painful if any package setups are incomplete

@angelathe great question, I incorrectly assumed the install command would fail but it sounds like it wouldn't and we do have some packages that run setup scripts. In order to allow those to run while still having `ignore-scripts=true` we'd have to use `@lavamoat/allow-scripts` which supports allow-listing packages to run their scripts and we'd also have to update our build scripts to support it. I will take a look into it if that seems reasonable.

angelathe:
> > Do we think there will be a lot of manual overhead using `ignore-scripts=true` (postinstall/setup steps)? I think we should include it, but slightly worried debugging could be painful if any package setups are incomplete
> 
> @angelathe great question, I incorrectly assumed the install command would fail but it sounds like it wouldn't and we do have some packages that run setup scripts. In order to allow those to run while still having `ignore-scripts=true` we'd have to use `@lavamoat/allow-scripts` which supports allow-listing packages to run their scripts and we'd also have to update our build scripts to support it. I will take a look into it if that seems reasonable.

@austin-hall-skylight @alismx Thoughts?

containers/ecr-viewer/README.md
```diff
@@ -94,7 +94,22 @@ By default, the seed data in the `star-wars` subfolder converts. To convert othe
 
 Additional commands can be found in [`package.json`](package.json).
 
-#### Creating a Migration
+### Adding packages
```
austin-hall-skylight, line 97: Thanks so much for including this!

austin-hall-skylight (approved):
🔒

PR #1483, opened 2026-05-21
fix: Adding more HAProxy env var configs

### LINKED ISSUE #1454: HAProxy Config - Additional environment variables
## What needs to be done

Make the following parts of the config use environment variables for ease of use:
- Number of instance slots inECS backend (set to 10 at the moment)
- Number of max connections
- Timeout queue
## Why it needs to be done

So that jurisdictions using ECS don't have to create a custom image to adjust those config values.

## Acceptance Criteria

- [ ] Update config to use environment variables and confirm it works 
- [ ] Update README file with new environment variable descriptions
- [ ] Update docker compose and TF files to set new env vars as needed

laura-puerto-skylight:
## Summary

Adding support for making more configs be driven by environment variables. I ended up including all timeouts just in case. 

Also added double quotes to env var wraps to make sure they are properly parsed.

## Related Issue

Fixes #1454 

## Acceptance Criteria

- [x]  Update config to use environment variables and confirm it works
- [x]  Update README file with new environment variable descriptions
- [ ]  Update docker compose and TF files to set new env vars as needed - not needed since we can set default values.

## Additional Information

## Checklist

### DIFF
--- a/containers/fhir-converter-proxy/haproxy.cfg
+++ b/containers/fhir-converter-proxy/haproxy.cfg
@@ -11,8 +11,8 @@ defaults
     # 10 seconds for AWS internal networking to connect
     timeout connect "${FHIR_CONVERTER_PROXY_CONNECT_TIMEOUT-10s}"
     
-    # Timeout while requests sit in the queue (defaults to 60s)
-    timeout queue "${FHIR_CONVERTER_PROXY_QUEUE_TIMEOUT-60s}"
+    # Timeout while requests sit in the queue (defaults to 300s)
+    timeout queue "${FHIR_CONVERTER_PROXY_QUEUE_TIMEOUT-300s}"
     
     # Increased timeouts are critical for large uploads over the internal network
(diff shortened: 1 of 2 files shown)


containers/fhir-converter-proxy/haproxy.cfg
```diff
@@ -9,16 +9,19 @@ defaults
     option httplog
     
     # 10 seconds for AWS internal networking to connect
-    timeout connect 10s
+    timeout connect "${FHIR_CONVERTER_PROXY_CONNECT_TIMEOUT-10s}"
+    
+    # Timeout while requests sit in the queue (defaults to 60s)
```
austin-hall-skylight: Is there any downside you can think of to increasing the default to 300s?

austin-hall-skylight (changes requested):

containers/fhir-converter-proxy/haproxy.cfg
laura-puerto-skylight: I don't think there should be, I'll update it!

austin-hall-skylight (approved):

PR #1522, opened 2026-06-10
feat: Return 409 for duplicate eCR in integrated viewer 

### LINKED ISSUE #1333: Return 409 for duplicate eCR in integrated viewer
## What needs to be done

Check blob storage and return a 409 if there is already a FHIR bundle with a name that matches the ID of the incoming eCR. This extends the functionality from #1332 to work with integrated configurations.

**Eng and prod to discuss whether to add a feature flag for this and what the default value should be**

## Why it needs to be done

We want to be able to return this error without having to process the eCR through the pipeline first to save time and compute.

## Acceptance Criteria

IF I send an eCR to `/process-ecr`
AND a FHIR bundle exists in blob storage with a name that matches the ID of the incoming eCR
THEN I will return a 409 without sending the eCR to the orchestration service

## Additional context

N/A

vochaparro2:
# PULL REQUEST

## Summary

* Updates the process eCR API to return a 409 when eCR is already loaded in the blob storage

## Related Issue

Fixes #1333 

## Acceptance Criteria

IF I send an eCR to /process-ecr
AND a FHIR bundle exists in blob storage with a name that matches the ID of the incoming eCR
THEN I will return a 409 without sending the eCR to the orchestration service

## Additional Information

## Checklist


containers/ecr-viewer/src/app/data/blobStorage/s3Client.ts
```diff
@@ -45,6 +46,25 @@ export const s3HealthCheck = async () => {
...
+  try {
+    await s3Client.send(
+      new HeadObjectCommand({
+        Bucket: process.env.ECR_BUCKET_NAME,
+        Key: objectKey,
+      }),
+    );
+    return true;
+  } catch {
+    return false;
```
austin-hall-skylight: I would recommend throwing an error instead of returning false here and in the cases where the GCP or Azure storage clients don't exist. Otherwise the caller is going to get a response that the issue was a duplicate eCR when it's actually an issue connecting to storage.

austin-hall-skylight (changes requested):

containers/ecr-viewer/src/app/api/save-fhir-data/service.ts
```diff
@@ -54,10 +63,19 @@ export const saveToStorage = async (
   }
 
   if (saveSource === S3_SOURCE) {
```
laura-puerto-skylight, line 72: Is there any chance that objectKey could be empty by the time we get here? Should we check and return an error if so?

containers/ecr-viewer/src/app/data/blobStorage/s3Client.ts
```diff
@@ -45,6 +46,25 @@ export const s3HealthCheck = async () => {
...
+ */
+export const existsInS3 = async (objectKey: string): Promise<boolean> => {
+  try {
+    await s3Client.send(
+      new HeadObjectCommand({
+        Bucket: process.env.ECR_BUCKET_NAME,
+        Key: objectKey,
+      }),
+    );
+    return true;
```
laura-puerto-skylight, line 66: I'm not familiar with the send() function but would it make sense to check what it returns before returning here? Not sure if it could return an error maybe. And if it's not necessary to check it maybe a comment explaining so would be helpful.

containers/ecr-viewer/src/app/api/save-fhir-data/service.ts
vochaparro2: Good catch, I've added a check for it.

containers/ecr-viewer/src/app/data/blobStorage/s3Client.ts
vochaparro2: I added a comment explaining why checking the return is not needed. Let me know if it needs more clarification.

containers/ecr-viewer/src/app/data/blobStorage/s3Client.ts
vochaparro2: Updated

laura-puerto-skylight (approved):
LGTM!

austin-hall-skylight (approved):

PR #1619, opened 2026-07-14
perf: batch lab inserts

### LINKED ISSUE #1511: SPIKE: eCR Viewer high CPU usage

austin-hall-skylight:
# PULL REQUEST

## Summary

Large eCRs can contain **thousands** of lab results meaning thousands of calls to the database to save metadata for a single eCR. This change batches lab record inserts to reduce database calls. SQL Server limits inserts to 4096 columns, each record we are inserting has 21 columns so the batch size is currently calculated to be 195 until we add or remove columns from ecr_labs.

## Related Issue

Fixes #1511

## Acceptance Criteria

- [ ] No change in resulting metadata saved
- [ ] Tests updated to test new behavior
- [ ] Improved eCR Viewer performance when ingesting eCRs

## Additional Information

## Checklist


laura-puerto-skylight (approved):
LGTM! Did you get a chance to test it with a really large ECR?

austin-hall-skylight:
@laura-puerto-skylight I did see about a 5 second gain (~6 seconds to ~1) in saving an eCR with about 3000 labs.

austin-hall-skylight:
Also want to flag that I reverted moving the max rows calculation out of the loop because the labs don't have the full number of columns at that point. I didn't want to hardcode the column count anywhere so I just switched it back to calculating after building the record.

PR #1621, opened 2026-07-15
fix: Updated condition list to aggregate conditions accross eCR versions

### LINKED ISSUE #1167: Multiple eCR versions with different RR conditions
## Description

- [ ] If an eCR has multiple versions and the versions have differing reportable conditions, all should be displayed.
- [ ] If there are no reportable conditions, the table should display "_No data_".

## Steps to reproduce

(No longer can reproduce because now `v3` shows COVID as a reportable condition, but this is how we originally found the bug)

1. View Mon Mothma versions of the eCRs (`v3` had no reportable conditions, `v2` and `v1` report COVID).
2. With no condition filter, Mon Mothma showed "no reportable condtions" for any version.
3. Filter for COVID -> Mon Mothma `v2` and `v1` were displayed, showing `COVID` as a reportable condition

## Priority

Lower priority primarily because we don't have example data for this anymore, but still probably an issue we'd like to fix eventually.

#### Priority

- [X] **Low priority** _The bug does not have to be fixed immediately. High and medium priority bugs should be addressed first._

## Notes

You could probably recreate this bug out by checking out a previous release version (i.e. `7.2.0`)

vochaparro2:
# PULL REQUEST

## Summary

  When a patient has multiple versions of an eCR, each version can carry different reportable conditions from the RR. Previously the eCR Library row only showed the conditions attached to the latest version, so conditions that appeared only on an  earlier version were invisible unless the user expanded and opened each related eCR.

  - The displayed row now shows the union of conditions from its own version **and                                                                                                                              
    all previous versions** in the eCR set joined through `set_id`                                                                                                                                                              
  - Conditions from *later* versions are never pulled in. This only comes into play                                                                                                                             
    when a condition filter matches an earlier version: that version is the one                                                                                                                                 
    displayed, and it shows only what was known as of that version.                                                                                                                                                                                                                                                                           
  - The conditions cell renders the standard "no data" placeholder when there are  no conditions                                                                                                                             


## Related Issue

Fixes #1167 

## Acceptance Criteria

## Additional Information

### Before:
### eCR would only show the Reportable Conditions and Rule Summaries for the most recent version:
**
<img width="1885" height="430" alt="image" src="https://github.com/user-attachments/assets/d61d551d-3199-4c16-b3d5-0d4093b778f1" />
**
### After:
### All conditions and rule summaries from previous versions are aggregated:
<img width="2148" height="870" alt="image" src="https://github.com/user-attachments/assets/8b9163c2-376e-4741-bd9f-40ad5de240c9" />


### Filtering by condition doesn't show reportable conditions for later versions:
<img width="2148" height="691" alt="image" src="https://github.com/user-attachments/assets/2501c549-02a7-440d-a228-b03b87f34683" />

## Checklist


laura-puerto-skylight:
It would be useful to have a before and after screenshot of the fix, do the rule summaries also need to be aggregated too @angelathe or @austin-hall-skylight ?

angelathe:
> It would be useful to have a before and after screenshot of the fix, do the rule summaries also need to be aggregated too @angelathe or @austin-hall-skylight ?

Yes to both! Rule summaries should also be aggregated (sorry about that Victor -- mentioned it in Slack but just saw that this wasn't mentioned in the original ticket 🤦‍♀️ ).

vochaparro2:
@angelathe @angelathe Thanks! I've updated it to aggregate rule summaries, updated the UI based on @echinelo-skylight feedback, and added screenshots 😄

echinelo-skylight (approved):
@vochaparro2 brilliant job with providing contextual screenshots. I really appreciate your inclusion of the filtering functionality accompanied with a screenshot too.

Again this was an excellent idea and it looks great!

laura-puerto-skylight (approved):
LGTM! Thanks for the screenshots!!

PR #1674, opened 2026-07-31
fix: Prevent FHIR evaluation cache collisions across eCR versions

angelathe:
# PULL REQUEST

## Summary

Problem: FHIR resource IDs can remain the same across different versions of an eCR even when the resource content changes. 

Examples: a preliminary lab result may become final, a patient may be marked as deceased....

- **Previously**: FHIRPath evaluation was caching (global) using resource ID and FHIR path.
    - Incorrect behavior: Viewer would return values cached from whichever version was evaluated first -> stale and/or incorrect FHIR information was appearing across bundles that shared those resource IDs (i.e. different eCR versions).
    - Ex. Mon Mothma v3 (Vital Status = Alive) was displaying Vital Status = Deceased when they were only marked as deceased in v4).

- **Solution**: Update evaluation cache to use the resource object as the key through a WeakMap. So even when FHIR resource IDs match, prevents values from one version leaking into another

Also updated tests (evaluation tests, snapshot tests that were prev. incorrect due to cache collisions)

## Screenshot
Mon Mothma v3 (Left) vs. Mon Mothma v4 (Right) -- Same patient resource ID, but no collision
Screenshot 2026-07-30 at 20 48 58

## Related Issue

Fixes # NA

Bug found during release `11.0.0` testing (though NOT introduced in this past release)

## Acceptance Criteria

## Additional Information

Possible cons: higher CPU? Open to any thoughts about improving this!

## Checklist

### DIFF
--- a/containers/ecr-viewer/src/app/utils/evaluate/index.ts
+++ b/containers/ecr-viewer/src/app/utils/evaluate/index.ts
@@ -2,7 +2,6 @@ import "server-only"; // fhirpath should only be used on the server
 
 import {
   Address,
-  Bundle,
   CodeableConcept,
   Coding,
   Element,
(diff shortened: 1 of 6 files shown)


vochaparro2 (approved):
LGTM! Good catch!

PR #1717, opened 2026-08-17
fix: Add Anatomical Region field to lab report section

### LINKED ISSUE #98: Add Anatomical Region field to lab report section
## What needs to be done

- Add new `Anatomical Region` field to lab 

## Acceptance Criteria

- [ ] Anatomical Region field is added above Anatomical Location/Laterality
- [ ] Anatomical Region displays `No data` when it is not available

## Technical Notes

- Anatomical Region is located in labReportJson
- An example of Anatomical Region can be found in Maine 2

vochaparro2:
# PULL REQUEST

## Summary

 Added Anatomical Region field to lab report section, but found a bigger issue while doing that. `getJsonLab` matches a `DiagnosticReport` to its narrative HTML table using an extension on the report's observations, but some results never carry that extension, so the match failed silently and every narrative sourced field on those reports showed no data, not just anatomical region. Added a fallback to the DiagnosticReport's own identifier when no observation has a usable reference value, since it carries the same numeric id the narrative uses.

## Related Issue

Fixes #98 

## Acceptance Criteria

- [ ] Anatomical Region field is added above Anatomical Location/Laterality
-- This wasn't possible since Anatomical Location is at the specimen level now and this field is at the Lab Report level.
- [x] Anatomical Region displays `No data` when it is not available

### Before
image

### After
image

## Checklist


containers/ecr-viewer/src/app/view-data/services/labsService.tsx
austin-hall-skylight: I know this isn't new to your change but this drew my attention to it. Could we get into any problems with partial ID matches due to using includes here instead of matching?

vochaparro2: Good catch! I made some changes to the logic. I think it's the best we can to do since we can't really do an exact match. It will now instead make sure the string ends with that number.

echinelo-skylight (approved):
Looks good. Thank you!

austin-hall-skylight (approved):

